PR #282, opened 2021-01-03
fields argument for the excel export

stumpyfr:
* `poetry run strictdoc export ~/Desktop/apex3-4.sdoc --formats=html,excel --fields=uid,statement,owner,priority,parent` is now working.
* unknown fields will create an empty column
* ran black on the modified files.


stumpyfr:
if you approve this idea, I can check to fix these unit tests.

stanislaw:
All looks good. Please proceed with writing one or more integration tests.

Check out the `invoke --list` command.

You mostly need `invoke test` to run all checks.

I think I have only written an integration test, not unit tests for Excel export. If you can a good idea of how to unit test this code as well, please feel free to add them as well. (`invoke test-unit`).

For integration tests:

You can use `invoke test-integration --focus 01_basic_excel_export` to run just one LIT test.

The idea there is that the `expected` folder contains the file you expect to be generated. So for the new export feature with special fields, you need to run the export, copy the file to the `expected` folder, otherwise your test should follow the same template as `strictdoc/tests/integration/commands/export/excel/01_basic_excel_export`.

stumpyfr:
Done.
for the integration/E2E tests, maybe worth a look to https://behave.readthedocs.io/en/stable/
Pretty fan of this package to easily describe "a la BDD" the features and test them.

stanislaw:
I have checked out your branch and all works and looks good.

Could you please `git rebase -i` your branch on top of the latest master and also squash your commits into just one?

I am generally following the following convention for the commit titles:

`feature: optional header: description`

In this case, something like `export/excel: export special fields` would work.

stanislaw:
@stumpyfr great work! Thanks for rebasing. One itest that could be written later would be to test exporting all possible fields to Excel. But that's another story.

stanislaw:
Another heads up: GitHub still does not show you as a contributor. It seems to be the behavior of the GitHub merging strategy when it is set to `Rebase` when merging the PRs.

I will use `merge` strategy from now on. Then your commits are yours and merge commits would be mine. (now it is single commit that has us both as co-authors).

Again, in case if this matters to you a lot, I didn't know about this behavior of GitHub either and from now on I will use the `merge` strategy instead of `rebase`.

stumpyfr:
no problem :)

stanislaw:
> Done.
> for the integration/E2E tests, maybe worth a look to https://behave.readthedocs.io/en/stable/
> Pretty fan of this package to easily describe "a la BDD" the features and test them.

I have worked with the BDD style tests myself but in the context of testing web applications.

Applying it to the command line applications is something that I have not considered before. Also, I didn't know about this specific tool. I am going to check it out. Thanks for the hint.

PR #314, opened 2021-01-21
parallelizer: fix hanging and zombie processes when child processes misbehave

### LINKED ISSUE #313: Parallelizing with multiprocessing can leave Zombie processes running
Currently, it is only reproducible by running `invoke test-integration` and killing the test process randomly.

stanislaw:
Python Multiprocessing: Crash in subprocess?
https://stackoverflow.com/questions/24894682/python-multiprocessing-crash-in-subprocess/24894997#24894997

Closes #313

### DIFF
--- a/strictdoc/helpers/parallelizer.py
+++ b/strictdoc/helpers/parallelizer.py
@@ -1,14 +1,9 @@
+import concurrent
 import multiprocessing
-import sys
+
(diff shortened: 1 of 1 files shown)


stanislaw:
The `OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES` on macOS might come back, but for now it seems to work ok.

PR #316, opened 2021-01-22
docs: performance feature highlight

stanislaw:


stanislaw:
cc @r-o-b-e-r-t-o, fyi: remember your question: `Questions on scalability of doorstop`? I am building StrictDoc along the topics that we discussed there.

Back then, I was impressed by your requirement of `hundred thousands to millions`. To this day, at my work I have only seen `a few hundreds, few thousands reqs max` reqs and specs documents and at that not-so-large scale, StrictDoc seems to work just fine.

Curious to hear back from you: have you found a tool that works for you since then?

PR #411, opened 2021-07-11
@mettta export/html: source code coverage

mettta:


mettta:
@stanislaw please review new function in /export/html/renderers/link_renderer.py

Issue #443, opened 2021-08-03
Include mathjax in HTML template to allow for TeX rendering in exported HTML


lochsh:
It would be great to be able to write TeX in the .sdoc the way that Doorstop allows, see https://doorstop.readthedocs.io/en/latest/reference/item/#example-latex-like-math-expressions

I think this is as simple as including the MathJax scripts in the HTML template, see https://www.mathjax.org/#gettingstarted

Possibly python-markdown-math is required, as mentioned in the Doorstop docs above.

Alternatively, being able to provide your own Jinja template would solve this problem while being more flexible.

lochsh:
Closed by #444

stanislaw:
The new strictdoc's pip 0.0.12 has been released.

PR #444, opened 2021-08-03
Add Mathjax to base HTML template to allow for TeX rendering

lochsh:
See #443. Note that this does not handle escaping, so backslahes in TeX will need
to be double backslashed.

For example: `\mathbf` would need to be `\\mathbf`

An example document:
```
[DOCUMENT]
TITLE: Test mathjax usage

[SECTION]
TITLE: Test

[FREETEXT]
$$
\\mathbf{\\underline{k}}_{\\text{a}} =
\\mathbf{\\underline{i}}_{\\text{a}} \\times
\\mathbf{\\underline{j}}_{\\text{a}}
$$
[/FREETEXT]

[/SECTION]
```

which is now rendering for me as 
image

It would be nice to also not require the double backslashes.

I was going to put the MathJax include in the `base.jinja.html`, but I see that other scripts have been moved to individual docs, so followed suit.


lochsh:
Apologies, I had the wrong issue number in my commit message, so I've force pushed to update it.

stanislaw:
Hi @lochsh,

This is a very interesting addition. Only two comments:

- We try to keep everything local without going to the internet for now. Could you download this script and put it together with the existing js files. We are fine to keep in the tree.

- I am not sure if this plugin should be always included in the templates by default. We will merge it as is for now, but eventually, also we might consider having a config option to activate/deactivate it.

Thank you for your contribution!

lochsh:
Thanks @stanislaw, I understand and will download the MathJax script.

It makes sense that you'd want to only include this script when needed -- I believe Doorstop does this, from looking at the outputted HTML when I was using it.

lochsh:
I'm having some issues now that I've got the MathJax script locally, I'll convert this PR to a WIP while I sort them out.

lochsh:
The issue is I very naively assumed I could just download the .js file, but I need to download everything in this directory https://cdn.jsdelivr.net/npm/mathjax@3/es5/. I'll sort that out tomorrow :+1:

stanislaw:
> The issue is I very naively assumed I could just download the .js file, but I need to download everything in this directory https://cdn.jsdelivr.net/npm/mathjax@3/es5/. I'll sort that out tomorrow 👍

Sounds good. The preferred option is to have a single file for this plugin.

stanislaw:
> I'm having some issues now that I've got the MathJax script locally, I'll convert this PR to a WIP while I sort them out.
> 
> EDIT: I don't think I have permission to convert the PR to a WIP.

I have checked – I don't think I have control over the permissions. The PR without WIP is fine 😄

stanislaw:
@lochsh, just a small note: I have pushed a commit that deactivates the HTML markup test (for now) because the test doesn't like the extra markup produced by the Mathjax library.

Let me know if you want this to be merged as is or you want to add more changes.

lochsh:
Thanks @stanislaw , I do still want to make changes, sorry for not getting round to them yet. I hope to do so later today, or tomorrow at the latest.

lochsh:
Hi @stanislaw 

I've updated the PR to have MathJax locally, which includes font files that it requires. This is about 508 KB total.

If you find this change too cumbersome now that it involves having to store these fonts, I understand. In that case, perhaps we could try making the HTML export more flexible, perhaps allowing users to include their own HTML file in the export, where they could (for example) include scripts such as MathJax.

Let me know what you think. If you are happy with this, it's ready to merge from a functionality perspective, as far as I can tell.

lochsh:
Actually, I think we're still missing a license and copyright notice for MathJax. Have you got an existing convention for where you usually store that sort of thing for third-party code?

stanislaw:
Several comments:

- I have just pushed the config option `--enable-mathjax` to your branch. Please check if it works on your side.

- We don't have any convention for story the licenses for 3rd-party software :smile: I think adding the Apache 2 license file to the `strictdoc/export/html/_static_extra/mathjax` should be enough for now. If you have recommendations, they are also appreciated.

- Regarding the customization for the HTML templates: This is an interesting idea, but I cannot think of a straightforward implementation for now. Having the Mathjax support under the config option is definitely fine for now.

- Once you confirm that the `--enable-mathjax` works for you, I will also update the documentation (images and Mathjax) and release new Pip.

lochsh:
Thanks @stanislaw! I've added a license file for MathJax.

The `--enable-mathjax` option works if I delete my output directory. But if I already have an output directory that was generated without the `--enable-mathjax` entry, it doesn't regenerate the HTML files so they don't have the script included.

I think if this was documented it wouldn't be a problem -- it's up to you whether it's worth fixing.

stanislaw:
> Thanks @stanislaw! I've added a license file for MathJax.
> 
> The `--enable-mathjax` option works if I delete my output directory. But if I already have an output directory that was generated without the `--enable-mathjax` entry, it doesn't regenerate the HTML files so they don't have the script included.
> 
> I think if this was documented it wouldn't be a problem -- it's up to you whether it's worth fixing.

I believe you are hitting an undocumented feature that is only mentioned at the beginning of the documentation: 

> ... From the second run, only changed documents are regenerated. Further performance tuning should be possible.

You have probably noticed that you often see log messages like these:

```
Skip: Sandbox SDoc .......................................... 0.00s
Published: StrictDoc ........................................ 1.99s
```

For large documentation trees, this saves time when only the changed SDoc files are regenerated. The static assets are not covered by this feature, so unless you change something in your `.sdoc` files the generator will not generate the HTML with the Mathjax library included. Deleting the `output/` folder will, of course, make the generator to create and copy everything from scratch.

I still didn't decide how to finalize and document this implementation. Let me know if it causes much trouble.

PR #458, opened 2021-08-16
export/html: update source-file-view, use micro requirement blocks

mettta:
- micro requirement blocks, 
- make source-file menu use file-tree, 
- move toggle switch 'Show coverage' into Ranges menu block


stanislaw (commented):
:+1:

PR #572, opened 2022-01-02
add Dockerfile for dockerization of strictdoc

### LINKED ISSUE #562: Dockerized strictdoc
Would be great to have a dockerized version of strictdoc similar to https://github.com/doorstop-dev/docker-doorstop .

fkromer:
resolves  #562

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,7 @@
 
 FROM python:3.10-alpine
 
-RUN apk add --no-cache bash git openssh
+RUN apk add --no-cache bash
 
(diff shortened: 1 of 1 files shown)


fkromer:
The version of `strictdoc` is pinned to v0.0.18. It can be changed easily by adjusting the version in line `    pip install --no-cache-dir strictdoc==0.0.18` according to pip/pypi convention in the `Dockerfile` before building the container image.

stanislaw:
Could you also add a note to the documentation, something like `Installing and running StrictDoc using Docker`?

To generate the documentation locally, you need to run `invoke sphinx`.

PR #574, opened 2022-01-03
add snap package

fkromer:
The version is hard pinned to `0.0.18` but can easily changed in `snapcraft.yaml` line [`version: '0.0.18'`](https://github.com/fkromer/strictdoc/blob/snap/snap/snapcraft.yaml#L3) and [`      - strictdoc==0.0.18`](https://github.com/fkromer/strictdoc/blob/snap/snap/snapcraft.yaml#L21).


snap/README.md
```diff
@@ -0,0 +1,13 @@
+# Build
+
+Install `snapcraft`: `sudo apt install snapcraft`
+
+Changedir into `snap`: `cd snap`
+
+Build `snap/strictdoc_0.0.18_amd64.snap`: `snapcraft`
+
+## Installation
```
fkromer, line 9: As soon as PR is merged the installation can be simplified to:

```
wget https://github.com/strictdoc-project/strictdoc/raw/snap/snap/strictdoc_0.0.18_amd64.snap
sudo snap install strictdoc_*.snap --devmode --dangerous
```

stanislaw: Could you also provide the documentation in the `docs/strictdoc.sdoc`? Please also address this comment at the same time: https://github.com/strictdoc-project/strictdoc/pull/572#issuecomment-1003721895. This way, you will add two new installation methods to the documentation.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,21 @@
+name: strictdoc
```
stanislaw, line 1: Would it be possible to keep this `snap` folder not in the root folder but inside the `developer` folder (`developer/snap`) or this is required by snap's conventions?

Background: I would like to keep the root folder not too crowded if possible.

stanislaw (changes requested):

snap/snapcraft.yaml
fkromer: Sounds reasonable. Fixed.

snap/README.md
fkromer: I've added the snap installation part in this PR. I'll address #572 in a follow up PR to keep topics separated. Do you want squashed PRs or is it ok to leave the single commits of this PR as they are?

stanislaw: Please address the #572 separately. Generally, I would prefer 1 commit by 1 PR with everything rebased every time. But this one I will merge as is :+1:

stanislaw (approved):

stanislaw:
@fkromer I have just checked the link to the `.snap` package and I am getting 404.

```
wget https://github.com/strictdoc-project/strictdoc/raw/developer/snap/snap/strictdoc_0.0.18_amd64.snap
--2022-02-20 09:54:22--  https://github.com/strictdoc-project/strictdoc/raw/developer/snap/snap/strictdoc_0.0.18_amd64.snap
Resolving github.com (github.com)... 140.82.121.4
Connecting to github.com (github.com)|140.82.121.4|:443... connected.
HTTP request sent, awaiting response... 404 Not Found
2022-02-20 09:54:22 ERROR 404: Not Found.
```

Similar to the maintenance of the brew package, I think the process of building snap/brew should be automated, otherwise I, personally, cannot find time to maintain these packages. I also didn't realize until now that a single `.snap` package takes 17MB which means that with each release, the size of the Git repository will increase by that size because the file is binary. Some way has to be found to attach them to [releases](https://github.com/strictdoc-project/strictdoc/releases) but even that will be maintenance overhead for me since I have limited capability to maintain them if something breaks.

I would like to know if you would volunteer to maintain the Snap/Brew packages in separate repositories because I am rather inclined to removing the Snap content from this master repo.

fkromer:
@stanislaw As `brew` packages are supported on Linux as well there is no need to provide a `snap` package. I did not recognize that `snap` packages are that bloated. I totally agree to remove the packaging aspects from this repository. I'll create a repo for the `brew` packaging but skip `snap` packaging.

PR #584, opened 2022-01-10
cli: add "version" command

fkromer:


stanislaw (changes requested):
The change looks good. Only two long-term maintenance comments:

- Last week I have prepared the contribution guidelines. Could I ask you to start rebasing your MRs from now on as explained here: ?
- Could you add an integration test for a `version` command here: https://github.com/strictdoc-project/strictdoc/tree/main/tests/integration/commands? You can use the regex matching functionality of filecheck such as `{{\d+.\d+.\d+}}` to match the version message using this test as an example: https://github.com/mull-project/FileCheck.py/blob/main/tests/integration/tests/py-only/version/sample.itest.

stanislaw (approved):

PR #602, opened 2022-02-02
Capella null pointer exception

cbernt:
added support for new Cappella Tags to sdoc export. 
Tested and confirmed working with html export.


stanislaw:
@cbernt thanks for looking into this further. Would it be possible for you to add an integration test that exercises the code that you are adding here? The changeset of this PR already has a sample of an integration test, and you may also check how the other `.itest`s are written.

This command should let you run just one integration test that you are working on. Just copy a folder with some existing itest to a new one, call it `wip` and then use it like this:

```
invoke test-integration --focus wip
```

An integration test would help me to see what you are exporting and what you are expecting to get. Additionally, we could use this test for next iterations and refactoring before the Capella feature gets integrated to the main branch.

PR #616, opened 2022-02-17
Windows path

cbernt:
Python should support Linux style path under Windows. 
As it turns out, this is only the case as long as the path string does not mix them, which occurs in the current implementation because path strings are dynamically built from different sources (user input, windows libs, files,...).

### DIFF
--- a/strictdoc/core/document_meta.py
+++ b/strictdoc/core/document_meta.py
@@ -1,6 +1,7 @@
 from strictdoc.export.html.document_type import DocumentType
 import os
 
+
 class DocumentMeta:
     def __init__(  # pylint: disable=too-many-arguments
         self,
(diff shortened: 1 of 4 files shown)


stanislaw:
@cbernt let's get the tests to a green state. Could you run `invoke check` on this branch because that's what the CI task does?

It looks like you have some code formatting issues that can be fixed by `invoke check` automatically.

stanislaw:
@cbernt this looks like a great step towards the full portability, including Windows.

Could I ask you to create at least one itest that has an example of the case when "the path string does mix them"? I see that you are adding normpath in several places in the code - it would be great to support this change with integration tests.

If you are wondering where it will be a good place for adding a test - just add a folder something like `windows_support` in the `tests/integration` root folder and add one or more tests for these mixed path strings.

In addition to adding the tests, I would like to ask you to follow the contribution guideline and rebase your PR to just one clean commit (see https://github.com/strictdoc-project/strictdoc/blob/main/CONTRIBUTING.md#git-workflow). To not solve two or three problems at the same time, please first:

1) Get this branch to green/success state
2) Add an integration test(s)
3) Rebase your PR to just one commit.

stanislaw:
@cbernt in the just-released 0.0.25, there should be a good improvement of how the paths are handled.

I am not a Windows user but I think I managed to get to a good consistent handling of the paths in a Windows VM. There is also a new integration test that checks for a case when a requirement has a link to a file and that link has backward slashes.

I would like to close this unfinished PR in favour of the work that has been released. I would be happy to hear from you if you had time to check the new behavior.

PR #620, opened 2022-02-21
traceability: support linking Tex files to requirements 

cbernt:
requirement references can now be added to latex files as comments. e.g. "% [REQ-001]"


stanislaw:
Thanks for the contribution @cbernt. I have only moved your test to the closest location of the feature and rebased your commits to just one.

PR #629, opened 2022-03-21
new feature FRAGMENT_FROM_FILE

BenGardiner:
proposed in #604. This introduces the ability to include 'fragments' of strictdoc files. Dual-purposing `[SECTION]` as the include target _and_ a section-as-usual didn't seem possible in the grammar and `processor.py` structure so a `[FRAGMENT]` top-level directive for include files is proposed here.

PR #632, opened 2022-03-23
Draft: Code review: [FRAGMENT] improvements 

stanislaw:


strictdoc/backend/sdoc/processor.py
```diff
@@ -64,10 +67,30 @@ def process_section(self, section: Section):
...
+            self._resolve_parents(include)
+        self.parse_context.current_include_parent = include.parent
+
+        reader = SDIReader()
+        fragment = reader.read_from_file(
+            file_path=include.file, context=self.parse_context
+        )
+        assert isinstance(fragment, Fragment)
+
+        include.parent.free_texts.extend(fragment.free_texts)
```
stanislaw, line 90: I am not sure if this code is fully stable. Considering the case when a fragment is included somewhere between the sections like:

```
free text
section
section
...
fragment
...
section
section
```

One option to eliminate this case will be to exclude free text from `FRAGMENT` and only allow them within the sections.

stanislaw: I will solve it by excluding the free texts.

stanislaw:
> From my perspective (with the caveat that I am not a StrictDoc dev, just a user) the include files (fragments) are either A) a section xor B) contents (i.e. freetext/composite/requirements/sections etc.) .

We seem to have two semi-related topics: 1) whether to have an option A or B and 2) whether to support free texts at the fragments's top level.

## 1) Option A vs Option B

The aspect that I tried to change with my commit was to avoid having the promotion of a fragment into a section, i.e. to avoid the option A. In your commit, the Fragment class shares the same properties with a Section which makes it a pseudo-object which I would like to avoid. Instead, I am proposing a `FRAGMENT` which is simply a top-level object for a fragment that can group the `section_contents`.

(*) By the current design, the top-level entity of an .sdoc document is `DOCUMENT` and the next-level entity that groups everything else is a `SECTION`. Given that the FRAGMENT's level is lower than that of a `DOCUMENT`, the next level of grouping becomes `SECTION`, so in the current code review commit, the `INCLUDE_SECTION_FROM_FILE` gets replaced by the fragment's `section_contents`, i.e. everything but free texts.

## 2) Free texts

It's been some time that I wanted to change the grammar to only allow zero or ONE free text to be connected to the `[DOCUMENT]` and zero or ONE free text connected to the `[SECTION]` - simply haven't done it yet. My understanding is that in the formal specifications documents which StrictDoc is trying to model you would never have a free-standing `FREETEXT` but rather always have it connected to a `DOCUMENT` or to a `SECTION`.

If we have this in mind, then combined with the implication from (*) I would prefer to avoid having the free texts "freely" floating at the `FRAGMENT`'s top level. Instead, a `[FREETEXT]` would always appear as part of a `[SECTION]`.

----

Having this said, I would say that my preference would be to keep the current solution as in the code review commit. At least, we should keep the `FRAGMENT` as a top-level tag that designates a fragment without making it to have a TITLE and share any other properties with the SECTION classes.

Regarding the `FREETEXT` my preference would be to always keep zero or one free texts connected to a section but I would be open to considering your use case (if it really exists). I think it should be doable to     

```py
     include.parent.section_contents.extend(fragment.section_contents)
     # AND
     include.parent.free_texts.extend(fragment.free_texts)
```

even though it will be against how I understood the free texts so far.

Do you have or see a real use case for:

1) Having "freely floating" free texts that are not connected to a DOCUMENT or a SECTION? (current design: no)

2) Having more than one free texts on the same level of nesting? (current design in the grammar: yes, but I would like to remove it and make it "zero or one")

I am open for a discussion.

BenGardiner:
> By the current design, the top-level entity of an .sdoc document is DOCUMENT and the next-level entity that groups everything else is a SECTION.

> [...] It's been some time that I wanted to change the grammar to only allow zero or ONE free text to be connected to the [DOCUMENT] and zero or ONE free text connected to the [SECTION] 

Aha . Thank you, this is two pieces of important information

> We seem to have two semi-related topics: 1) whether to have an option A or B and 2) whether to support free texts at the fragments's top level.

You're absolutely right. Good catch. :)

---

# Freetexts

> Do you have or see a real use case for:

> 1. Having "freely floating" free texts that are not connected to a DOCUMENT or a SECTION? (current design: no)

> 2. Having more than one free texts on the same level of nesting? (current design in the grammar: yes, but I would like to remove it and make it "zero or one")

No on both counts. I agree that a single freetext per section is the right model as I see it also.

---

# A) vs B)

So I think that resolves the freetexts but what is captured here still violates the principle of least surprise, I think.

> At least, we should keep the FRAGMENT as a top-level tag that designates a fragment without making it to have a TITLE and share any other properties with the SECTION classes.

I recognize the design intention here. And fulfilling those intentions leads to the additional changes here which results in needing always a `[SECTION]` under a `[FRAGMENT]` to avoid the ugliness that would otherwise occur with FRAGMENTs having TITLEs et. al.

From a user perspective, however, since this is a valid .sdoc:

```
[DOCUMENT]
TITLE: Test Doc

[SECTION]
TITLE: Section 1

[REQUIREMENT]
STATEMENT: Sub requirement

[/SECTION]

[REQUIREMENT]
```

the include-target-files (proposed as FRAGMENTS but we could call them w/e) needing this structure:

```
[FRAGMENT]

[SECTION]
TITLE: BLAH

[REQUIREMENT]
STATEMENT: also BLAH

[/SECTION]
```

implies that this is also a valid include-target-file (an option B case):

```
[FRAGMENT]

[SECTION]
TITLE: BLAH

[REQUIREMENT]
STATEMENT: also BLAH

[/SECTION]

[REQUIREMENT]
```

... which actually works (after minor changes) in these additional changes. 

Maybe it either my misunderstanding of "By the current design, the top-level entity of an .sdoc document is DOCUMENT and the next-level entity that groups everything else is a SECTION." or me poorly communicating option B.

If the above document examples make sense to you as valid then I know what to do to get this PR into shape.

stanislaw:
> Maybe it either my misunderstanding of "By the current design, the top-level entity of an .sdoc document is DOCUMENT and the next-level entity that groups everything else is a SECTION." or me poorly communicating option B.

You are right! My wording was imprecise: the above sentence is valid but in addition to that: 1) on each level, the requirements can be floating on the same level with sections, 2) the free texts are rather glued to, i.e. they belong to, either the documents or sections (0 or 1 free text for each document/section).

> If the above document examples make sense to you as valid then I know what to do to get this PR into shape.

Yes, please proceed (a clear Yes to `implies that this is also a valid include-target-file (an option B case):`). It should be possible to keep the requirements on a top level of a fragment, without them to be nested under a section.

BenGardiner:
We're all set in #629 . This can be deleted now

stanislaw:
This has been implemented in https://github.com/strictdoc-project/strictdoc/pull/629.

PR #633, opened 2022-03-28
requirement: use multiline value for None values in meta fields

BenGardiner:


stanislaw (approved):
Good catch! 

(I didn't anticipate this use case - that a meta field would be used multi-line.)

PR #651, opened 2022-04-04
export/html: support sections without a level

### LINKED ISSUE #639: Unnumbered sections
The default `HTML export` functionality of `strictdoc` generates numbered sections.
Is it possible to define a section as _unnumbered_ in the .sdoc source file?

The idea is to include a section header without affecting the level of the nested elements (i.e. requirements), but keeping the section title in a dedicated row in DOC, TBL and TR/DTR views.

Such an option can be very useful when dealing with source requirements documents with inconsistent sectioning (such as some technical standards/normative), that need to be included in a custom project.

stanislaw:
Closes #639


gioboske (approved):

gioboske (changes requested):
Works as expected for removing the section numbering, but it leads to uncorrected numbering for the nested requirements (except the first one).
Refer to [this comment](https://github.com/strictdoc-project/strictdoc/issues/639#issuecomment-1088108628) in #639 for an MRE showing the unexpected behaviour.

PR #662, opened 2022-04-15
feature: import excel

BenGardiner:
as discussed in #634 . Hope this is useful


BenGardiner:
wow, ok `invoke check` here did not prepare me for the automated checks...

stanislaw:
> wow, ok `invoke check` here did not prepare me for the automated checks...

Yeah, two things have happened recently:

1) The dependencies were relaxed from ~= to `>=` which means the newer versions of PyLint give more issues.
2) The invoke tasks are now running inside virtual environments which is great but still venv can "leak" into finding the dependencies in the system if they are present there. I find this rather confusing but don't know a good solution for this yet.

BenGardiner:
> > wow, ok `invoke check` here did not prepare me for the automated checks...
> 
> Yeah, two things have happened recently:
> 
> 1. The dependencies were relaxed from `~=` to `>=` which means the newer versions of PyLint give more issues.
> 2. The invoke tasks are now running inside virtual environments which is great but still venv can "leak" into finding the dependencies in the system if they are present there. I find this rather confusing but don't know a good solution for this yet.

no worries. I think 🤞 it's ready now

BenGardiner:
🥳

stanislaw:
> 🥳

Almost 😄 

Note that the .sdoc documentation was split into several files. It looks like your local copy still has the no longer existing `strictdoc.sdoc`. Please add the documentation to the `strictdoc-1-user-manual.sdoc` and regenerate the docs accordingly.

strictdoc/core/actions/import_action.py
```diff
@@ -0,0 +1,30 @@
...
+from strictdoc.cli.cli_arg_parser import (
+    ImportCommandConfig,
+)
+from strictdoc.helpers.timing import timing_decorator
+
+
+class ImportAction:
+    @staticmethod
+    @timing_decorator("Import")
+    def do_import(import_config: ImportCommandConfig):
```
stanislaw, line 16: Please add a type annotation for the return type.

stanislaw (changes requested):
Please address this round of the comments. I will need to take a closer look at your changes, inspect the Excel import code and the cli parsing tomorrow.

strictdoc/backend/excel/import_/excel_to_sdoc_converter.py
```diff
@@ -0,0 +1,223 @@
+import os
+from typing import Optional, Any
+
+import defusedxml
```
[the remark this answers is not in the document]
BenGardiner: good question! as it turns, out: since xlrd doens't _do_ xlsx anymore: it doens't need any mitigations via defused xml. Gone!

strictdoc/core/actions/import_action.py
BenGardiner: -> None ? Done

docs/sphinx/source/strictdoc.rst
```diff
@@ -0,0 +1,2062 @@
+Introduction
```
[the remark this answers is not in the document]
BenGardiner: yeah, I messed up the doc step here worse than that even. I'll fix it

BenGardiner:
ok I'm bad at this 😅

stanislaw (approved):

strictdoc/backend/excel/import_/excel_to_sdoc_converter.py
```diff
@@ -231,7 +230,7 @@ def extend_requirement_with_extra_columns(
         return requirement
 
     @staticmethod
-    def parse(input_path, title=None):
+    def parse(input_path):
```
BenGardiner: @stanislaw, can this commit be reverted please? We use the title argument in our code in https://github.com/nmfta-repo/nmfta-vehicle_cybersecurity_requirements/pull/10

stanislaw: Sure, could you do it? I generally try to avoid optional arguments but in this case, it would be great if you added the explanation that the intended use case if for the `.convert` method to be called from external scripts. Note that I have removed the `parse` method to simplify the code structure - now there is only `convert`.

BenGardiner:
will do

PR #677, opened 2022-05-05
HTML export improvements

BenGardiner:
1. don't use [no title] when there is a UID (but still no title)
2. don't link to 'UID None' when there is a UID (but still no title)
3. render really long meta fields next to rationale and comments

example before:
image

example after:
image

### DIFF
--- a/strictdoc/backend/sdoc/models/requirement.py
+++ b/strictdoc/backend/sdoc/models/requirement.py
@@ -8,6 +8,8 @@ from strictdoc.backend.sdoc.models.document_grammar import (
 from strictdoc.backend.sdoc.models.node import Node
 from strictdoc.backend.sdoc.models.reference import Reference
 
+MULTILINE_WORD_THRESHOLD = 6
+
 
 class RequirementContext:
     def __init__(self):
@@ -233,13 +235,30 @@ class Requirement(Node):  # pylint: disable=too-many-instance-attributes
             for single_field in requirement_field:
                 yield single_field
 
-    def enumerate_meta_fields(self):
+    def enumerate_meta_fields(
+        self, skip_single_lines=False, skip_multi_lines=False
+    ):
         for field in self.fields:
             if field.field_name in RESERVED_NON_META_FIELDS:
                 continue
-            meta_field_value = field.field_value
-            if meta_field_value is None:
-                meta_field_value = field.field_value_multiline
+            meta_field_value = (
+                field.field_value
+                if field.field_value
+                else field.field_value_multiline
+            )
+            if (
+                len(meta_field_value.splitlines()) > 1
+                or len(meta_field_value.split(" ")) > MULTILINE_WORD_THRESHOLD
+            ):
+                is_single_line_field = False
+            else:
+                is_single_line_field = True
+
+            if is_single_line_field and skip_single_lines:
+                continue
+            if (not is_single_line_field) and skip_multi_lines:
+                continue
+
             yield field.field_name, meta_field_value
 
     def dump_fields(self):
--- a/strictdoc/export/html/_static/_requirement.css
+++ b/strictdoc/export/html/_static/_requirement.css
@@ -278,6 +278,21 @@ dl.requirement_meta {
   background-color: var(--color-bg-main);
 }
 
+/* meta multiline list (next to comments/rationale) */
+
+.requirement_meta_multiline {
+  font-size: .9em;
+  font-style: italic;
+}
+
+.requirement_meta_multiline + .requirement_meta_multiline {
+  margin-top: var(--base-margin);
+}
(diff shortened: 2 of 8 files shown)


stanislaw:
Very good! Thanks a lot!

BenGardiner:
Awesome;

@stanislaw fyi it's rendering on main branch now https://nmfta-repo.github.io/vcr-experiment/vcr-experiment/00_tsrm.html

stanislaw:
@BenGardiner glad to hear it is working for you!

stanislaw:
This weekend I will release a new pip.

stanislaw:
A small follow-up on this one:

@BenGardiner, could I ask you about some feedback on how you use the TBL view in your work?

@mettta and I are looking into doing a small re-design of this screen because we want to solve a number of issues:

1) One issue is that a 3-column design very often results in just a few lines in one column (most often - center column) and many more lines in the right or left column. Your second screenshot is an example of this issue where your right column has much more content than the central one.
2) The meta information in the left column very often gets cut visually when viewed on smaller screens. On your first screenshot, the `:` of `VERIFICATION` is on the next line but we also saw other similar issues with our documents.

One idea we have is that for each requirement we could make the meta information be presented in a top row that spans over three columns. This would give enough width for all meta information to be presented without any visual cuts. After that the 3-column content would still follow like they are now: left - parent/child links, central - statement, right - comments, rationale and other longer fields.

Another idea we had would be to make it an Excel-like table extending right with horizontal scrolling to make sure that each field gets its own column but we are not sure if it is worth trying to implement Excel in HTML.

Another question is then whether you find it useful in your case that while a statement is only a few lines, your `PUB_REF` field is quite verbose and eats much more vertical space of the whole row.

The question is whether this TBL is useful at all compared to the DOC screen. Could you tell us more about your workflow - do you usually work with the DOC screen or TBL screen?

Also, just FYI and not related to this comment, we have introduced a document-wide option called `REQUIREMENT_STYLE`, see https://strictdoc.readthedocs.io/en/latest/strictdoc-html/docs/strictdoc-2-development-plan.html and the documentation: https://strictdoc.readthedocs.io/en/latest/strictdoc-1-user-manual.html#option-requirement-style. It only affects the presentation on the DOC screen.

BenGardiner:
> A small follow-up on this one:
> 
> @BenGardiner, could I ask you about some feedback on how you use the TBL view in your work?
> 
> @mettta and I are looking into doing a small re-design of this screen because we want to solve a number of issues:

Yes absolutely, thank you for checking with us.

> 1. One issue is that a 3-column design very often results in just a few lines in one column (most often - center column) and many more lines in the right or left column. Your second screenshot is an example of this issue where your right column has much more content than the central one.
> 2. The meta information in the left column very often gets cut visually when viewed on smaller screens. On your first screenshot, the `:` of `VERIFICATION` is on the next line but we also saw other similar issues with our documents.
> 
> One idea we have is that for each requirement we could make the meta information be presented in a top row that spans over three columns. This would give enough width for all meta information to be presented without any visual cuts. After that the 3-column content would still follow like they are now: left - parent/child links, central - statement, right - comments, rationale and other longer fields.

I think this will be fine.

> Another idea we had would be to make it an Excel-like table extending right with horizontal scrolling to make sure that each field gets its own column but we are not sure if it is worth trying to implement Excel in HTML.

Yes. That is probably not worth the effort IMHO.

> Another question is then whether you find it useful in your case that while a statement is only a few lines, your `PUB_REF` field is quite verbose and eats much more vertical space of the whole row.

Neutral.

> The question is whether this TBL is useful at all compared to the DOC screen. Could you tell us more about your workflow - do you usually work with the DOC screen or TBL screen?

 We don't have an explicit need for TBL view I think. The 00_tsrm.sdoc file you have been looking at is meant to replace this excel file https://github.com/nmfta-repo/nmfta-telematics_security_requirements/blob/master/Telematics%20Cybersecurity%20Requirements%20Matrix.xlsx . In that excel file is a 'printable matrix' (below) that is used in a PDF form of the matrix because some organizations need a published 'document' rather than an excel file. So pretty much any pretty-printed form of the requirements would be fine.

image

The other main deliverable from the matrix is the questionnaires tab -- this is a format I plan to add to strictdoc or as an external tool importing the module (whichever you find most palatable to the project).

image

But the TBL view per se is not key. 

> Also, just FYI and not related to this comment, we have introduced a document-wide option called `REQUIREMENT_STYLE`, see https://strictdoc.readthedocs.io/en/latest/strictdoc-html/docs/strictdoc-2-development-plan.html and the documentation: https://strictdoc.readthedocs.io/en/latest/strictdoc-1-user-manual.html#option-requirement-style. It only affects the presentation on the DOC screen.
 
❤️

PR #749, opened 2022-11-28
Trivial refactoring of has/get_children_requirements and has/get_pare…

GGBeer:
Separated part 2 Trivial refactoring of has/get_children_requirements and has/get_parent_requirements methods
See Discussions https://github.com/strictdoc-project/strictdoc/discussions/741 GrammarElementField-TypeValue and External-Refs

### DIFF
--- a/strictdoc/core/traceability_index.py
+++ b/strictdoc/core/traceability_index.py
@@ -1,3 +1,5 @@
+from typing import Dict
+
 from strictdoc.backend.sdoc.models.requirement import Requirement
 from strictdoc.backend.source_file_syntax.reader import (
     SourceFileTraceabilityInfo,
@@ -10,7 +12,7 @@ class TraceabilityIndex:  # pylint: disable=too-many-public-methods, too-many-in
     def __init__(  # pylint: disable=too-many-arguments
         self,
         document_iterators,
-        requirements_parents,
+        requirements_parents: Dict[str, Dict],
         tags_map,
         document_parents_map,
         document_children_map,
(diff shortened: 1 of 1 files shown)


stanislaw:
Now, let's consider this very different one separately. I am uncertain if I can agree that this refactoring is an improvement.

You are reducing the amount of duplication at the cost of creating a new artificial array for doing a trivial check of `has_child/parent_requirements`. Note that the `has_` methods only check None/emptiness and return False, in your code—you really have to access or create arrays to later check their `len()`.

The duplicated codes in these functions are intentionally like they are now. There is a certain amount of duplication between these functions, and they could be refactored to a little better state, but in a different way. I would like to suggest a small refactoring right on top of your change now.

GGBeer:
Depends on how costly a len(new [ ] ) instead of just return False is in this case. Returning None in the get_xxx() is no option as this breaks/changes its interface.
Which change do you have in mind?

stanislaw:
I have pushed my change. The main concern is of course not the performance, but simply the logic - there is no need to create an empty array to do the checks of has_ methods.

This code was written in a few iterations, and with time it got a little too defensive. I am comfortable having it the way it is for now.

I prefer to merge it this way, unless there was something special in your work that made you to refactor it your way. Anything that I might have missed?

I have revisited this data structure recently, and I think it needs to be changed to a data class or something better than a plain dictionary, but that should be another refactoring story:

```
                d_02_requirements_map[node.uid] = {
                    "document": document,
                    "requirement": node,
                    "parents": [],
                    "parents_uids": [],
                    "children": [],
                }
```

GGBeer:
In line 75 I added "not requirement.uid or" as I had created a requirement.uid==None case.

```
get_children_requirements(
if not requirement.uid or len(requirement.uid) == 0:

```
I will check when when I am going through the TypeValue branch.

stanislaw:
> In line 75 I added "not requirement.uid or" as I had created a requirement.uid==None case.
> 
> ```
> get_children_requirements(
> if not requirement.uid or len(requirement.uid) == 0:
> ```
> 
> I will check when when I am going through the TypeValue branch.

Shouldn't the `not requirement.uid` be covered by the line just above?

```
        if not isinstance(requirement.uid, str):
            return []
```

?

GGBeer:
Yes, it should be covered.

PR #772, opened 2022-12-12
tests/integration: try a Windows-friendlier diff command

stanislaw:


stanislaw:
@GGBeer it seems to work. I don't remember at this point what made me to use a custom Python script, but that was definitely Windows. Good to have it natively now.

PR #773, opened 2022-12-12
Code climate: Do not rely on SD working dirname in ConfTest

GGBeer:
Renaming StrictDoc's base Dirnname breaks pytest and itests.
Trying if conftest.py with only `../../..` works on all platforms.

### DIFF
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -1,9 +1,7 @@
 import os
 import sys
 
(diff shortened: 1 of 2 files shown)


stanislaw (approved):

stanislaw:
I have relaxed the settings a little: now you don't have to always rebase to the latest state. However, my preference of 1 rebased commit per 1 PR stays the same 😄

PR #788, opened 2022-12-19
export, rst: add custom grammar fields into jinja2 rendering

### LINKED ISSUE #779: export/rst: doens't render the custom fields
we are trying to use --formats=rst export in https://github.com/nmfta-repo/vcr-experiment to create a latexpdf.  We found that many of the fields in the grammar there 

```
[GRAMMAR]
ELEMENTS:
- TAG: REQUIREMENT
  FIELDS:
  - TITLE: UID
    TYPE: String
    REQUIRED: True
  - TITLE: LEVEL
    TYPE: String
    REQUIRED: False
  - TITLE: STATUS
    TYPE: String
    REQUIRED: False
  - TITLE: REFS
    TYPE: String
    REQUIRED: False
  - TITLE: TITLE
    TYPE: String
    REQUIRED: False
  - TITLE: STATEMENT
    TYPE: String
    REQUIRED: False
  - TITLE: TAGS
    TYPE: String
    REQUIRED: False
  - TITLE: RATIONALE
    TYPE: String
    REQUIRED: False
  - TITLE: COMMENT
    TYPE: String
    REQUIRED: False
  - TITLE: PUB_REFS
    TYPE: String
    REQUIRED: False
  - TITLE: VERIFICATION
    TYPE: String
    REQUIRED: False
  - TITLE: CRITICALITY
    TYPE: String
    REQUIRED: False
```

aren't rendered in the rst file. i.e. CRITICALITY, VERIFICATION, PUB_REFS

Is there a configuration we can set to render these extra/custom fields?

BenGardiner:
should resolve #779

### DIFF
--- a/strictdoc/export/rst/templates/requirement.jinja.rst
+++ b/strictdoc/export/rst/templates/requirement.jinja.rst
@@ -6,14 +6,17 @@
 {{ _print_rst_header(requirement.title, requirement.ng_level) }}
 {% endif %}
 
-{%- if requirement.uid is not none %}
+{%- if requirement.has_meta > 0 %}
 .. list-table::
     :align: left
     :header-rows: 0
-
-    * - **UID:**
-      - {{requirement.uid}}
-{% endif %}
+{% if true -%}{# without this workaround Jinja eats too much or not enough whitespace. #}{%- endif %}
+    {%- for meta_field in requirement.enumerate_meta_fields(skip_multi_lines=True) %}
+    * - **{{meta_field[0]}}:**
+      - {{ meta_field[1] }}
+    {%- endfor -%}
+{%- endif %}
+{% if true -%}{# without this workaround Jinja eats too much or not enough whitespace. #}{%- endif %}
 
 {%- if requirement.statement is not none %}
 {{requirement.statement}}
@@ -21,9 +24,20 @@
 {{requirement.statement_multiline}}
 {% endif %}
 
+{%- if requirement.rationale or requirement.rationale_multiline %}
+**Rationale:** {{ requirement.get_rationale_single_or_multiline() }}
+{% endif %}
+
 {%- for comment in requirement.comments %}
 **Comment:** {{comment.get_comment()}}
 {% endfor %}
+
+{%- if requirement.has_meta %}
+  {%- for meta_field in requirement.enumerate_meta_fields(skip_single_lines=True) %}
+**{{meta_field[0]}}:**
+{{ meta_field[1] }}
+  {%- endfor %}
+{%- endif %}
 
 {%- set requirement_references = requirement.get_requirement_references() %}
 {%- if requirement_references|length > 0 %}
--- a/tests/integration/commands/export/rst/08_custom_field/expected/input.rst
+++ b/tests/integration/commands/export/rst/08_custom_field/expected/input.rst
@@ -1,0 +1,31 @@
+.. _A-1:
+
+.. list-table::
+    :align: left
+    :header-rows: 0
+
+    * - **UID:**
+      - A-1
+    * - **TAGS:**
+      - has_comment, has_rationale
+    * - **CRITICALITY:**
+      - High
+
+
+the foo must bar
+
+**Rationale:** If the foo did not bar then it might instead baz, which must be avoided at all costs.
+
+**Comment:** A bazzing foo is dangerous because of the uncontrolled range of interaction of the baz with other foos.
+
+**PUB_REFS:**
+References
+
+1
+
+2
+
+**VERIFICATION:**
+Very detailed verification step;
+
+With multiple paragraphs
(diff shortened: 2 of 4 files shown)


tests/integration/commands/export/rst/08_custom_field/expected/input.rst
```diff
@@ -0,0 +1,16 @@
...
+.. list-table::
+    :align: left
+    :header-rows: 0
+
+    * - **UID:**
+      - A-1
+
+    * - **META_TEST:**
+      - Yes
+
```
stanislaw, line 12: I know, it can be quite tricky to get right, but could you ensure that the generator always generates no more than 1 space between logical blocks? You might need to play around with jinja's `{%-` compared to `{%`. Try to follow how I am doing it in the rest of the `strictdoc/export/rst/templates/requirement.jinja.rst`.

strictdoc/export/rst/templates/requirement.jinja.rst
```diff
@@ -6,25 +6,38 @@
...
-
+{%- endif %}
 {%- if requirement.statement is not none %}
 {{requirement.statement}}
 {% elif requirement.statement_multiline is not none %}
 {{requirement.statement_multiline}}
 {% endif %}
 
+{%- if requirement.rationale or requirement.rationale_multiline -%}
+    **Rationale:** {{ renderer.render_requirement_rationale(requirement) }}
```
stanislaw, line 26: I haven't checked the rendered PDF for this, but are you sure you want Rationale to be treated like a quote block?

stanislaw (changes requested):
This is excellent! And thanks for adding the integration test right away. I didn't expect this to be this easy/fast. Let's just address the two issues I have highlighted, and we will be good to merge.

strictdoc/export/rst/templates/requirement.jinja.rst
stanislaw: I mean, shouldn't it start like `Comment:` from the beginning of the line?

BenGardiner: I'll change it to a normal paragraph, no worries

tests/integration/commands/export/rst/08_custom_field/expected/input.rst
BenGardiner: yeah... it is tricky. let me see if I can wrestle it

stanislaw: Please do. If it gets hard, just drop it and let me know, and I could do another round myself. You might have noticed that I had to introduce the `{%- if true` hack to create a non-eatable string.

BenGardiner: > Please do. If it gets hard, just drop it and let me know, and I could do another round myself. You might have noticed that I had to introduce the `{%- if true` hack to create a non-eatable empty line.

so... I tried (probably more than reasonable) to get the extra line after multiple single-line meta fields to go away. Everything I tried either a) got rid of that extra newline _and_ the single new line after a single sngle-line meta (e.g. UID only) or b) kept the single newline in the prev case but added the extra newline in multi single line meta tags...

BenGardiner:
with respect to the checks failing: those are all whitespace differences in the rst exports right? Can we change the integration test commands to e.g. `diff -burp` instead?

tests/integration/commands/export/rst/08_custom_field/expected/input.rst
stanislaw: > with respect to the checks failing: those are all whitespace differences in the rst exports right? Can we change the integration test commands to e.g. `diff -burp` instead?

I liked your idea of a `long-bearing comment`. Unfortunately, it doesn't remove the trailing whitespace compared to what my hack does. The rst integration tests are passing now as is.

BenGardiner:
🧠👍

stanislaw:
Could I ask you to rebase these 3 commits to just one with your original commit name: `export, rst: add custom grammar fields into jinja2 rendering`? I would do it myself, but then, as I learned a few months ago, my rebase will make your commit be authored by me and I want to keep your contribution intact.

BenGardiner:
> Could I ask you to rebase these 3 commits to just one with your original commit name: `export, rst: add custom grammar fields into jinja2 rendering`? I would do it myself, but then, as I learned a few months ago, my rebase will make your commit be authored by me and I want to keep your contribution intact.

Yep! done

stanislaw (approved):

stanislaw:
Meanwhile, I have checked why you had difficulties with that double newline and found that we are not stripping the custom multiline fields as it is done for the normal reserved fields, see here: https://github.com/strictdoc-project/strictdoc/blob/2e6164a2896c9b0ede2dccb0a8a9a9bea0e67b82/strictdoc/backend/sdoc/models/requirement.py#L169.

The extra line was coming for you from the multiline field, not from Jinja 🥲 

I will fix this tomorrow-ish.

PR #789, opened 2022-12-20
backend/sdoc/grammar: Implement REFS-BibReference and add Document-Bi…

GGBeer:
…bliography Grammar supporting BibTex (file) entries.
Use FileEntry for FileReference and Biliography-BibTex files. 
Modified internal xyzReference field names to match their type of use.

Note: 
There still a lot work to do here e.g. add tests, rendering of cited bibliography items, handle duplicate bib-entries, export the Bibliography to Excel, etc.

Maybe you can have a look at the changes and provide some feedback.

Thx,
Guus

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -18,5 +18,5 @@ python-datauri >= 0.2.9, == 0.*
 beautifulsoup4 >= 4.9.3, == 4.*
 pygments >= 2.10.0, == 2.*
 lxml >= 4.6.2, == 4.*
-reqif >= 0.0.18, == 0.*
+reqif >= 0.0.20, == 0.*
 xlrd >= 2.0.1, == 2.*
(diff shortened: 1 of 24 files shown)


stanislaw:
I have tried to check what your branch produces visually but got this error:

```
jinja2.exceptions.UndefinedError: 'strictdoc.backend.sdoc.models.reference.BibReference object' has no attribute 'ext_bib_entry'
```

The error is not a problem at all, I just could't visualize how the output looks like.

Your contribution is very welcome and the way you approach it looks almost mergeable right away. Like with the previous PR, I kindly ask you to extract the refactoring part of the existing code to a separate PR which will simplify the review of both: 1) existing but refactored and 2) new parts. I have some minor comments, but I prefer to work on them with smaller changesets.

Your changes in `strictdoc/backend/sdoc/models/reference.py` are precisely a suitable candidate for the first PR (minus the introduction of `BibReference`, of course). Please extract it and then we both come back to review the remaining second PR (I estimate that it will not require a further split anymore).

PR #792, opened 2022-12-22
backend/sdoc/grammar: Add REFS-BibReference

GGBeer:
2nd step: Add REFS BibReference variant
Requires pybtex for parsing/handling BibTex entries.

Updating the User Manuel, add support for a Document.Bibliography, Validation of Citations will be done in a later step.


GGBeer:
Yep. Normally there will be the CitationKey (either the BibRef's citekey value as is or transformed into the more usual [`<number>`] or [`<AlphaNum>`], optionally followed by additional info for defining which detail/section of the book is referenced here.
Combined with the `[Document].[Bibliography]` part, a Bibliography section could be generated, which then only contains the cited BibTex-Items, which are linked by the `[Requirement].[Refs].[BibRef].[Citation]` entries.

Just chose the NW-Arrow to differentiate the BibRefs from the Parent/Child references.
I think the exact way of rendering the Bib-References and Bibliography needs some further discussion.

PR #796, opened 2022-12-25
Code climate: introduce helpers/auto_described to print objects

stanislaw:
The @auto_described implementation saves the effort of writing the custom __str__() descriptions for every class. The generalized implementation should be good enough for now.


stanislaw:
@GGBeer FYI

PR #797, opened 2022-12-28
backend/excel/export: Support Excel Export of custom grammar fields

GGBeer:


tests/integration/backend/excel/export/04_custom_grammar_fields_to_export/test.itest
```diff
@@ -0,0 +1,5 @@
+RUN: %strictdoc export --formats=excel --fields=uid,owner --output-dir=Output "%S/input.sdoc"
```
stanislaw, line 1: I have checked the expected Excel file manually, and it looks like you are not actually testing most of the things that you have in the `input.sdoc`. Could you remove the `--fields` argument, regenerate the Excel file, make sure that it looks like as expected, and finally force-push your change?

stanislaw (changes requested):
The changeset looks like going to the right direction overall, but it would be great if you strengthened the integration tests.

tests/integration/backend/excel/export/04_custom_grammar_fields_to_export/test.itest
GGBeer: Added some new Export Excel ITests. Testing showed that the specials "Parent" and "Comments" needed some work.

GGBeer: The new PR failed the ITest "07_custom_grammar_fields_to_export_fields_all_special" in the "Periodic Integration Test" 
Seems that pybtex somewhre needs the Six package? Not sure where to include/fix this.

stanislaw: That part of the integration test has to do with a rather tricky setup of PyInstaller. I am hoping to take a look into this still today.

stanislaw (approved):

stanislaw:
Thank you again for the contribution! I am merging this as a good step forward. In the near future, I think, it would be great to discuss the improvements of the layout of the produced Excel table. I feel like we would rather make it be formatted with font sizes like a document and somehow move to the "1 field per 1 row" approach. I don't have a clear visual solution to this yet, but some improvement is definitely needed.

Furthermore, similar to ReqIF, it would be great to add the "roundtrip" integration tests - an .sdoc is exported to an Excel file, then imported back, and you get a 100% original .sdoc again. It's something that's been on my list for a few months but the time was not there 😞

PR #811, opened 2023-01-10
backend/sdoc/grammar: Add Document-Bibliography

GGBeer:
Adding a Document Bibliography item.
Note: Central definition of Bib-Items only. Rendering/Exporting the cited Bibliography items is not implemented yet!


strictdoc/backend/sdoc/models/document_bibliography.py
```diff
@@ -0,0 +1,47 @@
...
+                    for bib_entry in bibtex_db.entries:
+                        self.bib_db.add_entry(
+                            bib_entry, bibtex_db.entries[bib_entry]
+                        )
+
+        if bib_entries and len(bib_entries) > 0:
+            for entry in bib_entries:
+                self.bib_db.add_entry(entry.ref_cite, entry.bibtex_entry)
+
+    def __str__(self):
```
stanislaw, line 41: Here you can use the latest `@auto_described`. See other classes.

strictdoc/backend/sdoc/grammar/type_system.py
```diff
@@ -38,12 +38,12 @@
 ;
 
 FileEntry[noskipws]:
-  ('  FORMAT: ' file_format = FileEntryFormat '\n')?
-   '  VALUE: ' file_path = /.*$/ '\n'
+  (/[ \-]/ ' FORMAT: ' file_format = FileEntryFormat '\n')?
```
stanislaw, line 41: Just to make sure I understand this:

For FileEntry you always have this in your tests:

```
- TYPE: BibRef
  FORMAT: String
  VALUE: SampleCiteKeyStringRef-1, "The sample BibReference String-Format"
```

and for the `BibEntry` below you have this:

```
- FORMAT: String
  VALUE: RFC-0000, Testing only
```

Would it make sense to hardcode the order of the fields by removing the flexibility: so the first field always has `- ` hardcoded and the second field has a double-space `  `?

tests/unit/strictdoc/backend/sdoc/test_dsl_passthrough.py
```diff
@@ -22,6 +22,11 @@
 from strictdoc.backend.sdoc.writer import SDWriter
 
 
+@pytest.fixture(autouse=True)
```
stanislaw, line 25: I normally try to avoid such pytest hacks. Could you explain which problem is this solving and maybe we could find a less hacky way?

stanislaw (changes requested):

strictdoc/backend/sdoc/grammar/type_system.py
stanislaw: My context here: I would like to avoid creating opportunities for undefined behavior? What if both components are specified without `- `? In that case, the grammar will look strange but will still work - and that's something that I would like to avoid. Fixing the order of the fields should mostly solve such cases.

tests/unit/strictdoc/backend/sdoc/test_dsl_passthrough.py
GGBeer: The file lookup for the bibtex testfiles only worked for the PyCharm Pytest run _or_ the "invoke check" Pytest run.
The pytest root-dirs are somehow different here. 
The fixture changes the current working directory to the directory of the current test module. Maybe we can set autouse=False and annotate/change test_170_grammar_document_bibliography() accordingly.

Other solutions?

strictdoc/backend/sdoc/models/document_bibliography.py
GGBeer: OK

strictdoc/backend/sdoc/grammar/type_system.py
GGBeer: Wanted to reuse the FileEntry and BibEntry classes.
Maybe I can make the grammar regex parsing more explicit to avoid the undefined behavior.
Also have to add more testcases.

strictdoc/backend/sdoc/models/document_bibliography.py
```diff
@@ -0,0 +1,47 @@
...
+        self.bib_db = BibliographyData()
+
+        self.bib_files = bib_files
+        self.bib_entries = bib_entries
+
+        if bib_files and len(bib_files) > 0:
+            for bib_file in bib_files:
+                if bib_file.file_format == FileEntryFormat.BIBTEX:
+                    # Just ignore any non-Bibtex files!
+                    bibtex_db = parse_file(
```
stanislaw, line 29: Putting a MARKER here which I would refer to in my other comment.

tests/unit/strictdoc/backend/sdoc/test_dsl_passthrough.py
stanislaw: I have taken a closer look at your branch, and I think, I don't have a final solution but just a few thoughts in the direction:

In the MARKER comment: https://github.com/strictdoc-project/strictdoc/pull/811/files#r1066920663, I see that you are calling a function that further parses a file and you are doing that from a Model-type object which is constructed by TextX. My assumption that the TextX-based objects should be pure data containers manipulated from the outside – they should not access the file system and take on the parsing responsibilities.

This is a non-critical difference but normally I prefer to do a more advanced processing, which goes beyond the parsing performed by TextX, in the `strictdoc/backend/sdoc/processor.py`. Now, if you moved the post-processing parsing code to the processor.py, we have our next problem and that is: the underlying magic access to the current working directory of the program, and I think, it is a good time to make that assumption explicit by providing an optional argument to the SDReader(path_to_root: str = os.getcwd()). Now when that is done, in your unit test you would simply make it explicit by providing an argument to the test file folder, similar to how I am doing it in the other places:

```
PATH_TO_THIS_TEST_FOLDER = os.path.dirname(os.path.abspath(__file__))
```

I hope this is not too vaguely described, please let me know.

strictdoc/backend/sdoc/models/document_bibliography.py
```diff
@@ -0,0 +1,47 @@
...
+    ):
+        self.parent = parent
+        self.bib_db = BibliographyData()
+
+        self.bib_files = bib_files
+        self.bib_entries = bib_entries
+
+        if bib_files and len(bib_files) > 0:
+            for bib_file in bib_files:
+                if bib_file.file_format == FileEntryFormat.BIBTEX:
```
stanislaw, line 27: And another pedantic comment: Is the expectation here that we always have the `FileEntryFormat.BIBTEX` type? How about adding an assert just above and left-indenting the whole code block by removing the whole if condition? Or you expect some variability here in the future? The motive is to make it strict and catch strange behavior as early as possible.

GGBeer:
Hi Stanislaw,
sorry, have to take care of another project since January.
The StrictDoc project is still ongoing, just didnot find time to complete this Bibliography-PR.
Still have to move the BibTex Processing away from the TextX region into the processer.py backend.

Also had some fun with rebase and merging mixed up with the many updates on the project's main branch. I have to be more carefull next time. However, I was able to cherry-pick my last squashed commits and apply them on a new branch
[Add-Doc-Bibliography-2](https://github.com/GGBeer/strictdoc/tree/Add-Doc-Bibliography-2) so nothing was lost.

Maybe you can reject the current PR and I will create a new PR when I complete the rework of the Biblograpy processing.
Thx, Guus.

stanislaw:
> Hi Stanislaw, sorry, have to take care of another project since January. 

Yeah, it's always like that 👍 

> The StrictDoc project is still ongoing, just didnot find time to complete this Bibliography-PR. Still have to move the BibTex Processing away from the TextX region into the processer.py backend.

Having thought about this retrospectively, I think we could merge your original PR as-is and then you would handle the move to the processing part in the next PRs. This would already have been a step forward and you would have to deal with less scope in your mind when refactoring.

The idea behind the processing is quite simple: your model classes only get the raw information from TextX parser, and then in the dedicated processor functions, you calculate some extra information and write it to the extra fields that you extend you objects with. You probably noticed how I use `ng_` prefix to indicate that these are extra processed fields that do not originate directly from the SDoc grammar text. All this is done to ensure that the model objects are rather data containers, not processing objects, which helps to use and maintain them easier long-term. The feature itself is quite nicely described by TextX here: http://textx.github.io/textX/3.1/metamodel/#processors.

> Also had some fun with rebase and merging mixed up with the many updates on the project's main branch. I have to be more carefull next time. However, I was able to cherry-pick my last squashed commits and apply them on a new branch [Add-Doc-Bibliography-2](https://github.com/GGBeer/strictdoc/tree/Add-Doc-Bibliography-2) so nothing was lost.
> 
> Maybe you can reject the current PR and I will create a new PR when I complete the rework of the Biblograpy processing. Thx, Guus.

No worries, the PR can stay as long as the feature is not integrated. Feel free to use this one or open a new one - do what fits best.

stanislaw:
Hi @GGBeer , I would like to check in about your plans regarding this work. Do you think, you could bring it to a working state in the future? Are you still interested in this feature yourself?

stanislaw:
> Hi @GGBeer, I would like to check in about your plans regarding this work. Do you think, you could bring it to a working state in the future? Are you still interested in this feature yourself?

Hi again, now due to a migration to Python 3.12, I am hitting

```
tests/unit/tools/test_confluence_html_table_import.py:5: in <module>
    from tools.confluence_html_table_import import ConfluenceHTMLTableImport
tools/confluence_html_table_import.py:8: in <module>
    from strictdoc.backend.sdoc.document_reference import DocumentReference
strictdoc/backend/sdoc/document_reference.py:3: in <module>
    from strictdoc.backend.sdoc.models.document import Document
strictdoc/backend/sdoc/models/document.py:4: in <module>
    from strictdoc.backend.sdoc.models.document_grammar import DocumentGrammar
strictdoc/backend/sdoc/models/document_grammar.py:5: in <module>
    from strictdoc.backend.sdoc.models.type_system import (
strictdoc/backend/sdoc/models/type_system.py:3: in <module>
    from pybtex.database import Entry
build/tox/py312-check/lib/python3.12/site-packages/pybtex/database/__init__.py:44: in <module>
    from pybtex.plugin import find_plugin
build/tox/py312-check/lib/python3.12/site-packages/pybtex/plugin/__init__.py:26: in <module>
    import pkg_resources
E   ModuleNotFoundError: No module named 'pkg_resources'
```

Given that the feature has not been completed, I am hesitant to fix this issue myself. Instead, I am leaning towards removing the unfinished code from StrictDoc's codebase. Please confirm whether you are intending to resume this development or not.

Issue #893, opened 2023-02-22
docs: extract to a separate repository, switch to SDoc documentation by default


stanislaw:
When the documentation for Read The Docs is regenerated, we get very large diffs like this one: https://github.com/strictdoc-project/strictdoc/pull/892. It does not make sense to store the auto-generated content along with the codebase.

The documentation itself can stay within the main repository but the generated artefacts should be pushed to another one.

stanislaw:
https://github.com/peter-evans/repository-dispatch might be useful here.

stanislaw:
It turns out that StrictDoc's own documentation can be deployed to Read the Docs along with the Sphinx HTML/PDF docs.

There is no need for a separate repo for the time being. Closing this in favor of https://github.com/strictdoc-project/strictdoc/pull/2023.

PR #1000, opened 2023-03-10
export/html: do not render unique identifiers when static export

stanislaw:


stanislaw:
This is a yet another bit of progress towards #893.

PR #1075, opened 2023-03-24
tests/end2end: try reducing the warmup interval from 3 to 0

stanislaw:
This should reduce the job time to almost 5 minutes, given 95 tests.


stanislaw:
Everything seems to work, and I prefer to merge it and hit/troubleshoot possible issues in the future instead of being afraid and waiting extra 5 min for every test run.

PR #1160, opened 2023-05-17
export/html: Connect to websocket relative to current domain

richardbarlow:
If `strictdoc server` is being forwarded/tunnelled, it will not necessarily appear to be coming from the client that is configured in the strictdoc conf file. Since the client is already aware of the host/port/protocol that is being used to access the backend, it is possible to construct a suitable websocket URL.


stanislaw:
This is excellent! I am merging, and the new Pip should be released within 48 hours.

PR #1257, opened 2023-07-25
backend/sdoc: optimize grammar for some more performance

stanislaw:


BenGardiner:
Hi @stanislaw; I'm pretty sure the merge of this PR broke vcr-experiment builds where we have (many) custom fields; but the error message here https://github.com/nmfta-repo/vcr-experiment/actions/runs/5680462987/job/15394634009 indicates a failure on the REFS field which starts with a RST bullet.

stanislaw:
HI @BenGardiner ,

As always, thanks for reporting this quickly. 

I have checked the line being reported and it looks like you are starting it with a blank newline. Could you explain if it has to be done that way?

I understood from our previous interaction that you wanted to the grammar to allow the whitespace on the first line but not necessarily allow empty newlines themselves? You don't seem to have newlines in the similar cases below and above the failing line.

https://github.com/nmfta-repo/vcr-experiment/blob/main/01_gateways.sdoc#L153

BenGardiner:
It's our pleasure to use this tool and help however we can. 

It took me a bit to check but there is _one_ case where a single blank line in a multi-line text field is needed: a blank like is needed when the first element in the multi-line text field is a bullet.

stanislaw:
It took some time for me to realize that you were referring to the RST export, not to the StrictDoc HTML export.

I have checked the existing `strictdoc/export/rst/templates/requirement.jinja.rst` template, and I think a better way of enabling the bullets is to make a newline break after the field name. This way, you no longer need to hack the bullet case by using an empty line.

I am almost certain I would prefer to have it this way because I find it more consistent, but I want to check with you if you are fine about this small RST layout change. Please let me know.

https://github.com/strictdoc-project/strictdoc/pull/1262

You can use this to try it with your repository:

```
pip install -U --pre git+https://github.com/strictdoc-project/strictdoc.git@stanislaw/fix_grammar_for_rst_edge_case
```

BenGardiner:
Aha yes that's a good idea. Yes, please emit a newline break into the RST
at the start of multiline blocks

Best Regards,
Ben Gardiner

stanislaw:
I have merged the MR https://github.com/strictdoc-project/strictdoc/pull/1262. Now you can update your .sdoc file by removing those blank newlines. Sorry for the inconvenience caused, but it looks like we are doing the right thing :smile:

BenGardiner:
That's perfect and no worries. :)

Issue #1310, opened 2023-09-28
Implement relation types (e.g., refines, implements, verifies)


stanislaw:
The implementation has already started but tracking the open items with this issue:

- [x] Grammar: Validate that a requirement has a valid relation (existing type/role).
- [x] Update grammar: Adding relations with types.
- [x] Update grammar: Adding child relation.
- [x] Update grammar validation: Duplicate general relation of a given type. 
- [x] Update grammar validation: No relations.
- [x] Update grammar validation: Duplicate non-general relation of a given type.
- [x] Update requirement: add a parent link with a role.
- [x] Update requirement: remove a child link with a role.
- [x] Update requirement: add a child link with a role.
- [x] Update requirement: remove a child link with a role.
- [x] Update requirement: role are selected from grammar roles.
- [x] Update requirement: prevent cycles when saving the form.
- [x] Update requirement validation: prevent both parent and child links with the same UID/role.
- [x] Update requirement: Adding second role-based relation (edge case).
- [x] UI: grammar form relations CSS.
- [x] UI: requirements form relations CSS.
- [x] UI: requirements view CSS.
- [x] Double-check the removal of child links from the traceability graph.
- [x] Compatibility of grammars between documents when linking with relations.
- [x] https://github.com/strictdoc-project/strictdoc/pull/1336
- [x] Start to recognize the RELATIONS use on the document tree level.
- [x] Disallow a String-typed REFS in the grammar.
- [x] Disallow a String-typed REFS in the requirement.
- [x] `assert_form_has_no_parents` - update
- [x] Deprecate REFS: field in the middle of a requirement.
- [x] Document the RELATIONS.
- [x] Document the child relations caveats.
- [x] Create requirement: handle new relations.
- [x] Create requirement: regenerate correct documents when new relations are added.

stanislaw:
At least 90% of this feature has been released in https://github.com/strictdoc-project/strictdoc/releases/tag/0.0.45.

I am afraid that I missed one or a few edge cases but the scope of this work became so large that I prefer to ship it for early testing.

PR #1320, opened 2023-10-10
Requirement relations: good 80%+ of the overall implementation

stanislaw:
This is an implementation of all major points from #1310. The overall changeset has grown very large, so it is time to integrate it, even thought there are smaller items still to be implemented.


stanislaw:
The branch got very big already, so I prefer to merge it and keep finishing the smaller items with follow-up MRs.

The four failing jobs need separate fixes and are not critical for getting this merged.

PR #1324, opened 2023-10-12
finders, generators: added yaml reference finder

villanella:
Added yaml finder and lexer according to https://github.com/strictdoc-project/strictdoc/discussions/1323


stanislaw:
Ignore the PyInstaller and dead links jobs. They are failing expectedly and need separate fixes each.

stanislaw:
Thanks for contributing this! I will not release a new Pip for quite some time because several other things need to be finished first.

What I can do this evening, is to release an alpha version which can be installed with `pip install strictdoc --pre`. I will do that later today.

Issue #1042, opened 2023-03-20
Rename globally: References -> Relations


stanislaw:
...and preserve the existing behavior for some time.

stanislaw:
After a few months, deprecate REFS and completely switch to RELATIONS.

richardbarlow:
We're just updating to v0.0.46 in our workflow and have seen that we need to perform the REFS->RELATIONS change and move the RELATIONS field to be the last. This was very easy to do using strictdoc passthrough.

However, it seems that when this change was implemented in strictdoc it re-introduced the concept of a 'warning' that was eliminated in https://github.com/strictdoc-project/strictdoc/issues/451. If there is a desire to have a separate concept of warnings then once again it would be helpful to be able to treat them as errors and exit with a non-zero exit code so that we can ensure that we have no warnings present.

stanislaw:
Hi, thanks for asking! This case with a deprecation warning is a little special because I wanted to avoid breaking things right away. I know of at least two users who would prefer to not have things break suddenly.

The fact that you noticed the changes and migrated was pretty much what I was aiming for.

> This was very easy to do using strictdoc passthrough.

It is good to know that this workflow has worked well for you.

The plan was to let the old REFS work for 3-6 months and then remove the old behavior for good. Along the way, I was planning to notify a few people who I know could be affected, including you. One thing connected to this is that I am currently missing a communication channel which I could use to notify as many existing users as possible. Only some users are in Discord, some users only use emails, some users are here on GitHub. Given this situation, the best thing would probably be to introduce a blog and a mailing list newsletter at the same time but I am not there yet with setting all of that up.

richardbarlow:
Thanks for the explanation. Personally I would expect to find information relation to deprecation of features and breaking changes in the changelog/release notes for the project. The critical thing is that the items written in the changelog/release notes are written for the intended user (i.e. a person looking to update the software) and are not just a dump of all of the commit messages or issues closed. Ideally breaking changes (or upcoming breaking changes) would be clearly highlighted.

An example of the release notes for a piece of software that we use and regularly update (with help from the release notes) can be found here: https://docs.zephyrproject.org/latest/releases/release-notes-3.4.html. The changes are written up in a manner that is meaningful for a person updating the software, grouped together to allow whole sections to be skipped if they are irrelevant and the sections are approximately ordered by importance.

[...]

PR #1466, opened 2023-11-17
docs: introduce the Changelog document

stanislaw:
Also regenerate the auto-generated changelog and maintain both documents together from now on.


stanislaw:
cc @richardbarlow 

This should answer your comment some time ago: https://github.com/strictdoc-project/strictdoc/issues/1042#issuecomment-1779217543. It took some time for me to get through some other work and come up with my thoughts regarding this.

Previously I had used `github_changelog_generator` quite successfully. It generates the changelog automatically, and I didn't have to write anything from myself like you suggested. Some time ago I started using GitHub releases where there also was the `auto-generate release notes` option, and I used it for a while.

The truth is that StrictDoc still has a very small user community so I was not sure if I actually needed to make an effort and document all the changes I made. Now, with your feedback, I am thinking of doing both the high-level descriptions and the auto-generated Changelog at the same time. In this PR, I have started the high-level changelog document and documented everything that will be released with the next Pip release. Let me know if this is what you would expect with every new release. The autogenerated part I will also maintain in parallel, so that the commit-by-commit history is available.

I was thinking of doing the high-level changelog myself, and your feedback certainly helped to prioritize and start doing it sooner than later.

richardbarlow:
This is wonderful. Thank you so much! It will make it much easier for us to update strictdoc in our pipeline and, with it making it easier for us to see the headline new features, make us want to update more frequently too :)

Issue #1509, opened 2023-12-14
Introduce concept of custom views


vik378:
For complex requirements documents its pretty common to have multiple attributes that are maintain by different roles. Displaying them all at the same time and in the same view however may cause confusion and loss of focus. Therefore most of the RM tools implement in some way a concept o customized view.

Lets assume that by default - if no view is defined - we see all attributes rendered. But if the VIEWS section is defined it can act in a more advanced way

Here is a proposal / example of how that VIEWS section could look like

```
[VIEWS]
- VIEW_ID: DEFAULT
   TAGS:
      REQUIREMENT:
        VISIBLE_FIELDS:
         - UID
         - STATEMENT
         - RATIONALE
      ASSUMPTION:
         VISIBLE_FIELD:
         - UID
         - STATEMENT
 - VIEW_ID: COMPLIANCE_CHECK
    TAGS:
        REQUIREMENT:
          VISIBLE_FIELDS:
          - UID
          - STATEMENT
          - ANTICIPATED_COMPLIANCE
          - MEANS_OF_COMPLIANCE
          - MEANS_OF_VERIFICATION
    HIDDEN_TAGS:
      - ASSUMPTION
```

This would also enable us to render Specifications in HTML / pdf with only the data that we want to share with a supplier / customer / stakeholder - for example you may want to hide reasoning, assumptions, design decisions while exporting your requirements to a supplier.

stanislaw:
## Register all models correctly, ensure that the Python classes are activated

```
DOCUMENT_MODELS = [
    DocumentConfig,
    Document,
    DocumentGrammar,
    DocumentBibliography,
    DocumentView,
    GrammarElement,
    GrammarElementFieldString,
    GrammarElementFieldSingleChoice,
    GrammarElementFieldMultipleChoice,
    GrammarElementFieldTag,
    GrammarElementFieldReference,
    GrammarElementRelationBibtex,
    GrammarElementRelationParent,
    GrammarElementRelationChild,
    GrammarElementRelationFile,
]
```


## Register a processing function

```
    def get_default_processors(self):
        return {
            "Document": self.process_document,
            "DocumentConfig": self.process_document_config,
            "DocumentGrammar": self.process_document_grammar,
            "DocumentView": self.process_document_view,
            "Section": self.process_section,
            "FragmentFromFile": self.process_include,
            "CompositeRequirement": self.process_composite_requirement,
            "Requirement": self.process_requirement,
[...]

PR #1600, opened 2024-01-17
(feat) Introduce concept of custom views

dahbar:
#1509 
- [x] Introduce a grammar part where the `[VIEWS]...` is modeled. Write a unit test, following the patterns from `tests/unit/strictdoc/backend/sdoc/test_dsl_passthrough.py` and the nearby tests.

- [ ] Validations of the Document SDoc:
  - [ ]  Ensure that the Views do not refer to FIELDS that are not registered in a document GRAMMAR.

 - [ ] Ensure that the Views do not refer to FIELDS that are not registered in a document GRAMMAR.
 HTML templates: Extend the Document template's Requirement partial with a filter that renderers fields based on a given view.

 - [ ] For CLI interface: Introduce a parameter to the `export `command that allows to export all documents only with a given view.

- [ ] For Web interface: Introduce an "Edit views" button that opens a screen for editing the views, similar to the "Edit grammar".

 - [ ] Create a ViewsObjectForm Python class that represents the edited content, following the patterns of "RequirementFormObject".

### DIFF
--- a/strictdoc.toml
+++ b/strictdoc.toml
@@ -34,7 +34,8 @@ exclude_doc_paths = [
 include_source_paths = [
   "pyproject.toml",
   "tasks.py",
-  "strictdoc/**"
+  "strictdoc/**",
+  "tests/**"
 ]
 
 exclude_source_paths = [
--- a/strictdoc/core/file_tree.py
+++ b/strictdoc/core/file_tree.py
@@ -182,7 +182,6 @@ class FileFinder:
                     and d != "build"
                     and d != "output"
                     and d != "Output"
-                    and d != "tests"
                 )
             ]
             dirs.sort(key=alphanumeric_sort)
--- a/tests/integration/self_testing/commands/passthrough/01_hello_world/test.itest
+++ b/tests/integration/self_testing/commands/passthrough/01_hello_world/test.itest
@@ -1,4 +1,4 @@
-RUN: %strictdoc passthrough "%strictdoc_root/" --output-dir="%S/Output/"
+RUN: cd "%strictdoc_root/" && %strictdoc passthrough "%strictdoc_root/" --output-dir="%S/Output/"
 
 RUN: %diff "%strictdoc_root/docs/strictdoc_01_user_guide.sdoc" "%S/Output/docs/strictdoc_01_user_guide.sdoc"
 RUN: %diff "%strictdoc_root/docs/strictdoc_02_faq.sdoc" "%S/Output/docs/strictdoc_02_faq.sdoc"


strictdoc/backend/sdoc/models/document.py
```diff
@@ -28,10 +30,9 @@ def __init__(
         self.title: str = title
         self.reserved_title: str = title
         self.config: DocumentConfig = (
-            config
-            if config is not None
-            else DocumentConfig.default_config(self)
+            config if config is not None else DocumentConfig.default_config(self)
```
stanislaw, line 33: It does not look like you are changing anything here. This will probably go away when you run `invoke check`.

strictdoc/backend/sdoc/grammar/grammar.py
```diff
@@ -161,6 +162,36 @@
   )?
 ;
 
+DocumentView[noskipws]:
+  'DEFAULT_VIEW: ' default_view = SingleLineString '\n'
```
stanislaw, line 166: Now that I see it implemented this way, I realize that we should rather move the `DEFAULT_VIEW` to the DocumentConfig as follows. It will make the meaning of this field consistent with the other fields because it is actually an option.

```
[DOCUMENT]
TITLE: Requirements
UID: SDOC_REQ
OPTIONS:
  REQUIREMENT_STYLE: Inline
  DEFAULT_VIEW: PRINT_VIEW
VIEWS:
...
```

Simply make the `DEFAULT_VIEW` to be the last field of the document config part. Also this makes sense when we get to editing the document options in the UI and the default view will be co-located with the other document options.

(This was not crystal clear for me in the beginning, sorry for making you to make this after we agreed on the format in our meeting).

strictdoc/backend/sdoc/models/document_view.py
```diff
@@ -0,0 +1,29 @@
...
+@auto_described()
+class ViewElement:
+    def __init__(
+        self,
+        id: str,
+        tags: List[ViewElementTags],
+        hidden_tags: Optional[List[ViewElementHiddenTag]] = None,
+        name: Optional[str] = None,
+    ):
+        self.id = id
```
stanislaw, line 20: It is better not to use `id` because it is a Python built-in and some quality tools complain that you overlap with Python's id:

```
strictdoc-project/strictdoc (html) $ python
Python 3.7.15 (default, Dec 29 2022, 19:39:20)
[Clang 14.0.0 (clang-1400.0.29.202)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> id
<built-in function id>
```

In this case, `view_id` would do just fine.

stanislaw, line 20: Please include the type annotations for the fields too:

`self.view_id: str = view_id`

This extra verbosity pays off when the static analyzers can give you helpful hints if you happen to use these types incorrectly somewhere else in the code. The developer guide mentions this.

I will not comment on the similar issues below. Please double-check this for all `self.*` fields in the PR.

strictdoc/backend/sdoc/models/document_view.py
```diff
@@ -0,0 +1,29 @@
...
+        self.tags: List[ViewElementTags] = tags
+        self.hidden_tags: Optional[List[ViewElementHiddenTag]] = hidden_tags
+        self.name = name
+
+
+class DocumentView:
+    def __init__(self, default_view: str, views: List[ViewElement]):
+        self.default_view = default_view
```
stanislaw, line 28: As commented above, the `default_view` should go to `DocumentConfig`.

strictdoc/backend/sdoc/writer.py
```diff
@@ -130,6 +132,37 @@ def write(self, document: Document):
                     output += requirement_in_toc
                     output += "\n"
 
+        document_view: DocumentView = document.view
+        if document_view:
+            assert document_view.default_view is not None
```
stanislaw, line 137: Besides the fact that the `default_view` should go to the DefaultConfig` as per my other comment, this field should be optional and therefore this assert should be removed because we want to support the case when there is no default view specified.

stanislaw (changes requested):
Very good work. There are only some a few comments that you should be able to address easily.

PR #1617, opened 2024-01-24
feat: Add grammar for custom views

dahbar:


strictdoc/backend/sdoc/writer.py
```diff
@@ -130,8 +131,39 @@ def write(self, document: Document):
...
+                        if placement:
+                            output += f"      PLACEMENT: {placement}\n"
+                hidden_tags = view.hidden_tags
+                if len(hidden_tags) > 0:
+                    output += "  HIDDEN_TAGS:\n"
+                    for hidden_tag in hidden_tags:
+                        output += f"  - {hidden_tag.hidden_tag}\n"
+
         assert document.grammar is not None
-        document_grammar: DocumentGrammar = document.grammar
```
stanislaw, line 134: This change looks strange. Was there a reason to do it?

strictdoc/backend/sdoc/writer.py
```diff
@@ -130,8 +131,39 @@ def write(self, document: Document):
...
+                for tag in view.tags:
+                    output += f"  - OBJECT_TYPE: {tag.object_type}\n"
+                    output += "    VISIBLE_FIELDS:\n"
+                    for field in tag.visible_fields:
+                        output += f"    - NAME: {field.name}\n"
+                        placement = field.placement
+                        if placement:
+                            output += f"      PLACEMENT: {placement}\n"
+                hidden_tags = view.hidden_tags
+                if len(hidden_tags) > 0:
```
stanislaw, line 160: `hidden_tags` is declared as optional and should be checked for `hidden_tags is None and len(hidden_tags) > 0`. (I am wondering why mypy didn't complain about this :thinking: )

I cannot see exactly what textX does for this array but sometimes it always stored an empty array and sometimes a list depending on how a grammar is written.

Check for `is not None` is a safer way to go.

strictdoc/backend/sdoc/writer.py
```diff
@@ -130,8 +131,39 @@ def write(self, document: Document):
                     output += requirement_in_toc
                     output += "\n"
 
+                if default_view is not None:
+                    output += "  DEFAULT_VIEW: "
+                    output += default_view
+                    output += "\n"
+
+        document_view = document.view
+        if document_view:
```
stanislaw, line 140: Please use the full check `if document_view is not None`.

stanislaw (changes requested):

strictdoc/backend/sdoc/writer.py
dahbar: None at all, for some reason I thought this was my change.

stanislaw (approved):

PR #1629, opened 2024-01-29
feat: Add validation for views

dahbar:


stanislaw (approved):
The Views backend part is taking shape now. Great work!

PR #1638, opened 2024-02-07
export/html: implement --view option 

stanislaw:


strictdoc/export/html/templates/components/requirement/meta/index.jinja
```diff
@@ -11,7 +11,7 @@
 {% endif %}
 
 {%- if requirement.has_meta -%}
-  {% for meta_field in requirement.enumerate_meta_fields(skip_multi_lines=True) %}
+  {% for meta_field in requirement.enumerate_meta_fields(skip_multi_lines=True) if current_view.includes_field(requirement.requirement_type, meta_field[0]) %}
```
stanislaw, line 14: Would be great to see that this line and all other lines are stressed by the integration test: `01_view_option_is_provided`.

PR #1641, opened 2024-02-09
fix: Review changes

dahbar:


tests/integration/commands/export/options/--view/01_view_option_is_provided/test.itest
```diff
@@ -3,4 +3,7 @@ RUN: %cat %S/Output/html/01_view_option_is_provided/input.html | filecheck %s --
 
 CHECK-HTML: REQ-1
 CHECK-HTML: Requirement statement.
+CHECK-HTML: Requirement second extra field.
```
stanislaw, line 6: This is almost what we need. There is one limitation, however, for CHECK-NOT checks: they are only effective between a CHECK before and a CHECK after. This means that you better arrange the CHECK and CHECK-NOT just in the order how the grammar fields are defined. This way you really make sure that the CHECK-NOT are matched at the right time. You also could prove that this works for every check by temporarily disabling the `--view` option in the itest and seeing that the CHECK-NOT checks indeed get hit.

PR #1669, opened 2024-02-26
feat: Make the document grammar editable with respect to human titles

dahbar:
For an example document:

```
[DOCUMENT]
TITLE: TestDoc
OPTIONS:
  MARKUP: HTML

[GRAMMAR]
ELEMENTS:
- TAG: REQUIREMENT
  FIELDS:
  - TITLE: UID
    TYPE: String
    REQUIRED: False
  - TITLE: TITLE
    TYPE: String
    REQUIRED: False
  - TITLE: STATEMENT
    TYPE: String
    REQUIRED: False
  - TITLE: CUSTOM_FIELD
    HUMAN_TITLE: Custom Field
    TYPE: String
    REQUIRED: False
  RELATIONS:
  - TYPE: Parent

[REQUIREMENT]
UID: REQ-001
TITLE: Test Document
STATEMENT: Test Statement
CUSTOM_FIELD: Test Field
```

We have the following form:

image

We can edit the human title and after saving, the example document becomes:

```
[DOCUMENT]
TITLE: TestDoc
OPTIONS:
  MARKUP: HTML

[GRAMMAR]
ELEMENTS:
- TAG: REQUIREMENT
  FIELDS:
  - TITLE: UID
    TYPE: String
    REQUIRED: False
  - TITLE: TITLE
    TYPE: String
    REQUIRED: False
  - TITLE: STATEMENT
    TYPE: String
    REQUIRED: False
  - TITLE: CUSTOM_FIELD
    HUMAN_TITLE: Custom Field 2
    TYPE: String
    REQUIRED: False
  RELATIONS:
  - TYPE: Parent

[REQUIREMENT]
UID: REQ-001
TITLE: Test Document
STATEMENT: Test Statement
CUSTOM_FIELD: Test Field
```

### DIFF
--- a/strictdoc/export/html/form_objects/document_grammar_form_object.py
+++ b/strictdoc/export/html/form_objects/document_grammar_form_object.py
@@ -48,11 +48,13 @@ class GrammarFormField:
         self,
         field_mid: str,
         field_name: str,
+        field_human_title: Optional[str],
         field_required: bool,
         reserved: bool,
     ):
         self.field_mid: str = field_mid
         self.field_name: str = field_name
+        self.field_human_title: Optional[str] = field_human_title
         self.field_required: bool = field_required
         self.reserved: bool = reserved
 
@@ -62,12 +64,16 @@ class GrammarFormField:
         return GrammarFormField(
             field_mid=grammar_field.mid,
             field_name=grammar_field.title,
+            field_human_title=grammar_field.human_title,
             field_required=grammar_field.required,
             reserved=reserved,
         )
 
     def get_input_field_name(self):
         return f"document_grammar_field[{self.field_mid}][field_name]"
+
+    def get_input_field_human_title(self):
+        return f"document_grammar_field[{self.field_mid}][field_human_title]"
 
 
 @auto_described
@@ -136,9 +142,15 @@ class GrammarElementFormObject(ErrorObject):
         document_grammar_fields = request_form_dict["document_grammar_field"]
         for field_mid, field_dict in document_grammar_fields.items():
             field_name = field_dict["field_name"]
+            field_human_title = field_dict.get("field_human_title")
+            if field_human_title is not None:
+                field_human_title = field_human_title.strip()
+                if len(field_human_title) == 0:
+                    field_human_title = None
             form_object_field = GrammarFormField(
                 field_mid=field_mid,
                 field_name=field_name,
+                field_human_title=field_human_title,
                 field_required=False,
                 reserved=is_reserved_field(field_name),
             )
@@ -295,7 +307,7 @@ class GrammarElementFormObject(ErrorObject):
             grammar_field = GrammarElementFieldString(
                 parent=None,
                 title=field.field_name,
-                human_title=None,
+                human_title=field.field_human_title,
                 required="True" if field.field_required else "False",
             )
             grammar_fields.append(grammar_field)
@@ -395,6 +407,7 @@ class GrammarElementFormObject(ErrorObject):
         field: GrammarFormField = GrammarFormField(
             field_mid=MID.create(),
             field_name="",
+            field_human_title=None,
             field_required=False,
             reserved=False,
         )
--- a/strictdoc/export/html/form_objects/grammar_form_object.py
+++ b/strictdoc/export/html/form_objects/grammar_form_object.py
@@ -34,7 +34,11 @@ class GrammarElementFormField:
         self.document_mid: str = document_mid
 
     @staticmethod
-    def create(*, grammar_element: GrammarElement, document_mid: str):
+    def create(
+        *,
+        grammar_element: GrammarElement,
+        document_mid: str,
+    ):
         return GrammarElementFormField(
             field_mid=grammar_element.mid,
             field_name=grammar_element.tag,
--- a/strictdoc/export/html/templates/components/form/field/contenteditable/field_with_human_title.jinja
+++ b/strictdoc/export/html/templates/components/form/field/contenteditable/field_with_human_title.jinja
@@ -1,0 +1,71 @@
+{%- assert field_class_name is defined, "field_class_name is defined" -%}
+{%- assert field_editable is defined, "field_editable is defined" -%}
+{%- assert field_input_name is defined, "field_input_name is defined" -%}
+{%- assert field_label is defined, "field_label is defined" -%}
+{%- assert field_label is not none, "field_label is not none" -%}
+{# FIXME: Remove #}
+{%- assert field_name is not defined, "field_name is not defined" -%}
+{%- assert field_placeholder is defined, "field_placeholder is defined" -%}
+{%- assert field_placeholder is not none, "field_placeholder is not none" -%}
+{%- assert field_value is defined, "field_value is defined" -%}
+{%- assert field_human_title is defined, "field_human_title is defined" -%}
+{%- assert mid is defined, "mid is defined" -%}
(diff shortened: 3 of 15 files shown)


strictdoc/export/html/form_objects/document_grammar_form_object.py
```diff
@@ -136,11 +142,15 @@ def create_from_request(
         document_grammar_fields = request_form_dict["document_grammar_field"]
         for field_mid, field_dict in document_grammar_fields.items():
             field_name = field_dict["field_name"]
+            field_human_title = field_dict.get("field_human_title")
+            if field_human_title is not None and not field_human_title.strip():
```
stanislaw, line 146: Here, let's not do several things on one line. I would prefer it this way:

```py
if field_human_title is not None:
    field_human_title = field_human_title.strip()
    if len(field_human_title) == 0:
        field_human_title = None
```

Note that the `.strip` is important for both empty and non-empty string cases, and my example ensures that.

strictdoc/export/html/form_objects/document_grammar_form_object.py
```diff
@@ -50,11 +50,13 @@ def __init__(
         field_name: str,
         field_required: bool,
         reserved: bool,
+        field_human_title: Optional[str] = None,
     ):
         self.field_mid: str = field_mid
         self.field_name: str = field_name
         self.field_required: bool = field_required
         self.reserved: bool = reserved
+        self.field_human_title: Optional[str] = field_human_title
```
stanislaw, line 59: Please make the order of both arguments and fields as follows:

```
self.field_name: str = field_name
self.field_human_title: Optional[str] = field_human_title
self.field_required: bool
self.reserved: bool = reserved
```

This way you would maintain a better rhythm.

strictdoc/export/html/form_objects/grammar_form_object.py
```diff
@@ -28,22 +28,33 @@ def __init__(
...
         self.document_mid: str = document_mid
+        self.field_human_title: Optional[str] = field_human_title
 
     @staticmethod
-    def create(*, grammar_element: GrammarElement, document_mid: str):
+    def create(
+        *,
+        grammar_element: GrammarElement,
+        document_mid: str,
+        field_human_title: Optional[str] = None,
```
stanislaw, line 43: This `field_human_title` argument does not seem to be used anywhere. Let's remove it.

strictdoc/export/html/form_objects/grammar_form_object.py
```diff
@@ -28,22 +28,33 @@ def __init__(
...
+    ):
         return GrammarElementFormField(
             field_mid=grammar_element.mid,
             field_name=grammar_element.tag,
             document_mid=document_mid,
+            field_human_title=field_human_title,
```
stanislaw, line 49: Since this is a case when a new grammar element form field is created we are good to have it explicitly to None:

```
field_human_title=None
```

(StrictDoc's code style is rather verbose with the added value of a maximum transparency. The default parameters are almost never used.)

strictdoc/export/html/form_objects/grammar_form_object.py
```diff
@@ -28,22 +28,33 @@ def __init__(
         field_mid: str,
         field_name: str,
         document_mid: str,
+        field_human_title: Optional[str] = None,
```
stanislaw, line 31: Let's remove the default `= None` part.

(StrictDoc's code style is rather verbose with the added value of a maximum transparency. The default argument values are almost never used.)

strictdoc/export/html/form_objects/grammar_form_object.py
```diff
@@ -84,10 +95,14 @@ def create_from_request(
         ]
         for field_mid, field_dict in document_grammar_fields.items():
             field_name = field_dict["field_name"]
+            field_human_title = field_dict.get("field_human_title")
+            if field_human_title is not None and not field_human_title.strip():
```
stanislaw, line 99: See my other comment about `.strip` above. The same applies.

strictdoc/export/html/templates/components/form/field/contenteditable/field_with_human_title.jinja
```diff
@@ -0,0 +1,71 @@
+{%- assert field_class_name is defined, "field_class_name is defined" -%}
```
stanislaw, line 1: We will do this a little differently when @mettta adds a realistic UI. Good for now.

Probably the `field_with_human_title` will be removed. Instead, the `row_with_reserved_field` will contain two fields, one for name, one for human title. @mettta will decide on the way.

strictdoc/export/html/templates/components/form/field/contenteditable/index.jinja
```diff
@@ -31,21 +30,18 @@
     {%- endif -%}
     data-testid="form-field-{{ testid_postfix }}"
   >
-    {%- if true -%}
-      {{ field_value }}
-    {%- endif -%}
+    {{- field_value -}}
```
stanislaw, line 33: I didn't know about this one. Good solution that removes our `if true` hack.

strictdoc/export/html/templates/components/form/field/contenteditable/index.jinja
```diff
@@ -31,21 +30,18 @@
...
-  {%- if field_type == "multiline" -%}
-    <textarea hidden name="{{ field_input_name }}">
-      {%- if true -%}
-      {{ field_value }}
-      {%- endif -%}
-    </textarea>
-  {%- endif -%}
+    {%- if field_type == "multiline" -%}
+      <textarea hidden name="{{ field_input_name }}">
+        {%- if true -%}
```
stanislaw, line 41: So here you can use the same `{{- -}}` solution?

stanislaw (changes requested):
Only minor changes requested. Great work otherwise! Should be mergeable when the changes are addressed.

strictdoc/export/html/form_objects/document_grammar_form_object.py
dahbar: I thought the same about this, but changing the order runs into a pylint issue:

`error: Failed to parse strictdoc/export/html/form_objects/document_grammar_form_object.py:44:9: non-default argument follows default argument`

strictdoc/export/html/templates/components/form/field/contenteditable/field_with_human_title.jinja
dahbar: Sounds good! This was meant to provide the simplest UI possible with the corresponding functionality.

strictdoc/export/html/form_objects/document_grammar_form_object.py
stanislaw: Are you saying that you get this pylint issue if you have it like this:

```py
 def __init__(
        self,
        field_mid: str,
        field_name: str,
        field_human_title: Optional[str], 
        field_required: bool,
        reserved: bool,
    ):
```

?

dahbar: Yes! Actually the interpreter gives you an issue as well. 

I just found [this solution](https://dev.to/rajeshj3/solved-non-default-argument-follows-default-argument-400d) which I have never seen before introduced in python 3.4:


```python
def greet(*, name='Nicolas', place):
    return f"Welcome {name}, to {place}"
``` 

with an "*" at the beginning.

stanislaw (approved):
Great stuff! Thanks for taking time to address all changes!

PR #1720, opened 2024-03-20
Fix a crash when following the user guide.

MartyLake:
Strictly following the user guide for this part yield to a crash because the requirement number 2 does not exist.


stanislaw:
Thanks!

PR #1798, opened 2024-04-27
pickle: catch AttributeErrors on schema changes, raise AssertionError otherwise

stanislaw:


stanislaw:
Thanks @nashif for reporting this.

Issue #1766, opened 2024-04-12
Extend GUI to support adding FREETEXT blocks


gregshue:
I would like the GUI to support adding/editing FREETEXT blocks in the same way it supports adding sections and items. It would be even better if the default content for FREETEXT blocks was "TBD".

stanislaw:
Thanks for reporting this. Adding FREETEXT blocks using UI is on the roadmap and tracked here: https://github.com/strictdoc-project/strictdoc/issues/1518.

@mettta and I were planning to implement this still in 2024-Q2.

stanislaw:
Closed with the release: https://github.com/strictdoc-project/strictdoc/releases/tag/0.0.57. Please see the release notes.

Issue #1518, opened 2023-12-19
Issue with Mixing FREETEXT and REQUIREMENT in .sdoc Files


dahbar:
### Environment
StrictDoc Version:  0.0.47
Python Version: 3.11.6

### Description:
I've encountered an issue when working with .sdoc files in StrictDoc. The problem arises when I attempt to use a mix of FREETEXT and REQUIREMENT elements in a single document. It would be great if the document allowed alternating between these two types of tags, but currently, this does not seem to be supported.

### Steps to Reproduce:

1. Create a .sdoc file.

2. Insert the following content:
```
[DOCUMENT]
TITLE: TestDoc

[GRAMMAR]
ELEMENTS:
- TAG: REQUIREMENT
  FIELDS:
  - TITLE: UID
    TYPE: String
    REQUIRED: True
  - TITLE: TITLE
    TYPE: String
    REQUIRED: True
  - TITLE: STATEMENT
    TYPE: String
    REQUIRED: False

[FREETEXT]
Test
[/FREETEXT]

[REQUIREMENT]
UID: TEST-0111
TITLE: T1
STATEMENT: Test

[FREETEXT]
Test
[/FREETEXT]

```
3. Try to run the StrictDoc server

### Error
```
raise self.nm
arpeggio.NoMatch: Expected Not at position ../strictdoc/TestDoc.sdoc:(30, 2) => 'T: Test  [*FREETEXT] '.
```

Was this intended or is there something I am missing?

stanislaw:
One thing that could help us during the implementation:

So far I haven't seen a document where a free text would be between the surrounding requirements or a free text would be the last one in a section or a document. Could you give an example of your immediate use case?

In other words, I would like to understand which kind of information requires these in-between free texts? It doesn't have to be your real work example but something that gives me an idea. Thanks.

rosensymmetri:
I ran into the same issue while manually translating a pretty unstructured text document into sdoc. Something like this:

> This may not happen under one of the following conditions:
> 1. Situation
> 2. Situation
>
> Short sidebar discussion about 2
>
> 3. Situation
>
> Longer discussion about all of the above

I wanted to translate the three numbered items into requirements, and the other items into freetexts.

dahbar:
Same use case for me. Unstructured documents some type of information in the middle between requirements or a footer with extra notes about the document.

```
[DOCUMENT]
TITLE: TestDoc

[GRAMMAR]
ELEMENTS:
[...]

PR #1833, opened 2024-05-29
backend/sdoc: grammar: teach SDocNodeField to parse FreeText parts

stanislaw:
This is a very good step towards #1766 and #1518.

### DIFF
--- a/strictdoc/backend/excel/import_/excel_to_sdoc_converter.py
+++ b/strictdoc/backend/excel/import_/excel_to_sdoc_converter.py
@@ -14,6 +14,7 @@ from strictdoc.backend.sdoc.models.node import SDocNode, SDocNodeField
 from strictdoc.backend.sdoc.models.object_factory import SDocObjectFactory
 from strictdoc.backend.sdoc.models.reference import ParentReqReference
 from strictdoc.backend.sdoc.models.type_system import GrammarElementFieldString
+from strictdoc.helpers.string import ensure_newline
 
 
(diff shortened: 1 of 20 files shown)


haxtibal:
@stanislaw This is great, thanks!

We can now almost link to arbitrary elements from any other elements content. The missing pieces seem rather small, see my demo implementation in commit e1a7979. Did you have that already on your agenda? If not, should I try a PR?

stanislaw:
Yes, this change was important, and it makes your `Demo...` change much easier. 

Note that LINK and ANCHOR have been surprisingly rich with all sorts of edge cases that need careful handling.

It is easy to introduce your change and it would work for static HTML which generates the TraceabilityIndex only once. With the editable UI flow, several further things have to be handled: what if a requirement is removed, and somewhere there is a link that points to it? Or what if a requirement changes its UID (which normally should not happen but this case has to be considered).

This group of end2end tests should give you an example of what is done for Section and ANCHOR: https://github.com/strictdoc-project/strictdoc/tree/main/tests/end2end/screens/document/_cross_cutting/anchors/section. Also, looking for [LINK and [ANCHOR in the existing integration and end2end tests should show you a complete picture of the changes needed.

stanislaw:
One comment though: I think we could merge your Demo commit without making all of the UI changes right away. It can just be documented in the existing limitations in the docs and the UI supports browsing but does not support editing requirements with [LINK] and [ANCHOR].

PR #1835, opened 2024-05-30
Feature: LINK to requirements and custom nodes (static export only for now)

haxtibal:
Currently, a `[LINK: ID]` tag is limited to target either sections or anchors. This change extends the feature so that we can link to `[REQUIREMENT]`s or any other element from a custom grammar. E.g. the following will work now:

```
[DOCUMENT]
TITLE: Document

[GRAMMAR]
ELEMENTS:
- TAG: CUSTOM
  FIELDS:
  - TITLE: UID
    TYPE: String
    REQUIRED: False
  - TITLE: DESCRIPTION
    TYPE: String
    REQUIRED: True

[FREETEXT]
[LINK: CUSTOM-1]
[/FREETEXT]

[CUSTOM]
UID: CUSTOM-1
DESCRIPTION: foo
```

The logic to create the link title in case of requirement or custom element is:
1. Use title field, if defined.
2. Else use UID.
3. Elements without UID are not relevant, as one needs an an UID string to write a LINK tag in the first place.

For cleaner design, we introduce a small wrapper class `LinkableNode` that centralizes rules for how to create the link title.

### DIFF
--- a/docs/strictdoc_01_user_guide.sdoc
+++ b/docs/strictdoc_01_user_guide.sdoc
@@ -1654,14 +1654,16 @@ UID: SDOC_UG_LINKS_AND_ANCHORS
 TITLE: Links
 
 [FREETEXT]
-StrictDoc supports creating inline links to document sections and anchors.
+StrictDoc supports creating inline links to document sections, anchors, requirements and custom grammar elements.
 [/FREETEXT]
 
 [SECTION]
-TITLE: Section links
+TITLE: Links
 
 [FREETEXT]
-When a section has an UID, it is possible to reference this section from any other section's text using a ``[LINK: <Section UID>]`` tag.
+Elements that have an UID can be referenced from section text using a ``[LINK: <UID>]`` tag.
+To reference a section that has an UID, use ``[LINK: <Section UID>]`` tag.
+Likewise, a requirement can be referenced with ``[LINK: <Requirement UID>]``.
 
 Example:
 
--- a/strictdoc/core/traceability_index.py
+++ b/strictdoc/core/traceability_index.py
@@ -78,6 +78,29 @@ class GraphLinkType(IntEnum):
     DOCUMENT_TO_TAGS = 7
 
 
+class LinkableNode:
+    def __init__(self, node: Union[SDocNode, SDocSection, Anchor]):
+        self.inner_node = node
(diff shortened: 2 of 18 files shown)


haxtibal:
Hi @stanislaw, I intend to create a few more tests, but would be interested in your general opinion first.

stanislaw:
This is a very good step forward, and it is been quite a while that I wanted to do this myself. Please proceed with the tests you were going to add.

haxtibal:
@stanislaw From your comments in https://github.com/strictdoc-project/strictdoc/issues/1833#issuecomment-

> what if a requirement is removed, and somewhere there is a link that points to it? Or what if a requirement changes its UID

We have the same potential for inconsistency here:
- Let a freetext contain a link to a requirement. Then a web UI user removes that requirement.
- Let a freetext contain a link to a requirement. Then a web UI user changes the UID of that requirement.

I started to handle it, but am currently stuck on how to handle web dialog interactions:
1. On click to requirement -> delete, server sends a http 200 turbo-stream answer "Deleting a requirement is an unrecoverable action". The UI displays this with a sdoc-modal element.
2. On click to "Delete Requirement", my validation kicks in server side, detects there's a reference, and causes a http 422 response.
3. The "Deleting a requirement is an unrecoverable action" doesn't go away. How to make it disappear and show the validation error instead?

I can probably figure this out on my own but it will take some time for learning. If you know you could help me out with some pointers.

BTW: There's a similar situation when a linked-to section is deleted in the UI. With strictdoc 0.0.55 it is currently unhandled. The section is removed, no error is reported, the referreing freetext has a dangling LINK afterwards. Is this known/intended?

stanislaw:
I have merged this PR as a good start for static export, and I have opened a follow-up ticket here and copied in your comments: https://github.com/strictdoc-project/strictdoc/issues/1839.

> BTW: There's a similar situation when a linked-to section is deleted in the UI.

Good catch. This is not known and is something that I missed before. I have added it to the list of items on the follow-up ticket.

Thank you for contributing this initial patch!

I am ready to complete the remaining parts by myself but I first need to finish implementing the `[TEXT]` behavior. I guess, you are not depending on the editable web interface anyway, so for you, the lack of these validations in the UI is fine for now?

PR #1855, opened 2024-06-07
html: Rename "parent relations" to "parents" and "child relations" to "children"

haxtibal:
This should improve comprehensibility and is consistent to what the rst generator already does.

Also add a sentence in the user documentation to explain by example who in the relation is the parent and who has the active role.


stanislaw:
Thanks!

Issue #1876, opened 2024-06-19
Bug: UI requirement validation fails if optional SingleChoice field is not entered


haxtibal:
To reproduce, start `strictdoc server` for the following empty document and go to the web UI.

```
[DOCUMENT]
TITLE: Overview

[GRAMMAR]
ELEMENTS:
- TAG: REQUIREMENT
  FIELDS:
  - TITLE: UID
    TYPE: String
    REQUIRED: True
  - TITLE: TITLE
    TYPE: String
    REQUIRED: False
  - TITLE: STATEMENT
    TYPE: String
    REQUIRED: True
  - TITLE: MAYBE
    TYPE: SingleChoice(A, B, C, D)
    REQUIRED: False
```

Then try to add a new requirement
- enter text in fields UID, TITLE, STATEMENT
- leave field MAYBE blank

Expectation: UI validation passes. It's ok to leave MAYBE blank, because it has `REQUIRED: False` in the grammar.

Actual: UI validation fails with "Node's MAYBE must be a value one of A, B, C, D. ".

stanislaw:
Thanks for reporting this as well!

I was too short-sighted when rushing to implement this quickly. Another part missing is the validation of MultipleChoice fields which I wanted to complete still in Q2.

Fixing this should be rather easy.

stanislaw:
I have released a new nightly Pip with a fix (`pip install --pre`).

The rest of the related work is tracked here: https://github.com/strictdoc-project/strictdoc/issues/1076. If editing MultipleChoice fields is important for you right now, consider contributing a PR with the same kind of validation.

(I am working towards finishing the FREETEXT-TEXT epic here: https://github.com/strictdoc-project/strictdoc/issues/1864, and most of the must-parts are in place now. Hoping to release a new version by the beginning of next week. After that, there will be a usual re-prioritization of everything, and something else will be worked on next.)

Issue #1076, opened 2023-03-25
UI: Grammar with non-string types


stanislaw:
Also editing requirement with non-string grammar types:

- [x] SingleChoice: Allow editing options in the UI using a dropdown menu.
- [x] SingleChoice: Prevent a node/requirement form from being saved if a value is out of range.
- [x] MultipleChoice: Prevent a node/requirement form from being saved if a value is out of range.

```
NotImplementedError: GrammarElementFieldSingleChoice(title = "COVERAGE_SINGLE", options = [2 elements], required = False, _tx_position = 357, _tx_position_end = 440, parent = GrammarElement(...), gef_type = "SingleChoice")
```

This ticket can be done together with https://github.com/strictdoc-project/strictdoc/issues/1022.

The follow-up work: #2230.

johanenglund:
Dropdowns in UI for single and multi choice is given a +1 upvote by me!

stanislaw:
Closing this in favor of only one follow-up issue: https://github.com/strictdoc-project/strictdoc/issues/1076.

PR #1878, opened 2024-06-20
UI: Support MultipleChoice in requirement form

haxtibal:
As suggested yesterday in #1876.

### DIFF
--- a/strictdoc/export/html/form_objects/requirement_form_object.py
+++ b/strictdoc/export/html/form_objects/requirement_form_object.py
@@ -2,7 +2,7 @@
 import html
 from collections import defaultdict
 from enum import Enum
-from typing import Dict, List, Optional, Set
+from typing import Dict, List, Optional, Set, Union
 
 from starlette.datastructures import FormData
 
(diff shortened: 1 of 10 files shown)


haxtibal:
The form takes raw strings, is that ok for now? I mean, one could instead want a drop down in the UI from which you select available choices, and a "+" to add the next choice.

As a consequence, I added sanitizing separators, so that if you enter "A,B" it is silently changed to "A, B" with the exact ", " separator as the parser wants it. Is that ok?

stanislaw:
> The form takes raw strings, is that ok for now? I mean, one could instead want a drop down in the UI from which you select available choices, and a "+" to add the next choice.

This is exactly what the issue https://github.com/strictdoc-project/strictdoc/issues/1076 is tracking. By having this PR opened, you are addressing one of the three items. For now, the raw strings are exactly ok.

> As a consequence, I added sanitizing separators, so that if you enter "A,B" it is silently changed to "A, B" with the exact ", " separator as the parser wants it. Is that ok?

That is definitely ok.

Thanks for doing this 👍

Issue #1897, opened 2024-06-24
[LINK: <uid>] does not allow to express all valid UIDs


haxtibal:
Currently there are two different sets of allowed characters for a UID, depending on the UID context:
```
[DOCUMENT]
TITLE: Document UIDs must not contain dots or whitespaces.
UID: FOO_BAR_1

[REQUIREMENT]
UID: FOO BAR 2.
TITLE: Requirement UIDs may contain dots or whitespaces.
STATEMENT: Test.
```

That's because DOCUMENT, SECTION, LINK, ANCHOR syntax use `REGEX_UID = r"([A-Za-z0-9]+[A-Za-z0-9_\-]*)"` in their grammar, while requirements define UID as SDocNodeField which most often is matched with regex `\S.*`. [StrictDoc User Guide - UID](https://strictdoc.readthedocs.io/en/stable/strictdoc_01_user_guide.html#uid) confirms the latter:

> StrictDoc does not impose any limitations on the format of a UID. Examples: ... SAVOIR.OBC.PM.80

As a consequence, one can define `FOO BAR 2.`, but will get an exception when trying to reference it with `[LINK: FOO BAR 2.]`.

IMO:
- any UID that can be defined should also be usable in LINKs
- allowed UID characters should be the same in any context, for consistency and comprehensibility
- allowing whitespaces and dots is useful when copying requirements from external documents/standards to sdoc

stanislaw:
You are catching many things that I wanted to do myself for a long time, these are great reminders :+1: 

Your IMOs are also how I see it.

stanislaw:
Do you think you could create a good regex to catch all the cases and reuse it everywhere in the grammar? A few unit tests in the `passthrough*` group would be nice to have.

haxtibal:
> Do you think you could create a good regex to catch all the cases and reuse it everywhere in the grammar?

Yes, should be doable, just can't promise when I get to it.

haxtibal:
My initial thought was to have a special line for UID, just as it's already done for MID, something like
```
SDocNode[noskipws]:
  '[' !'SECTION' !SDocCompositeNodeTagName requirement_type = RequirementType ']' '\n'
  ('MID: ' mid = SingleLineString '\n')?
  ('UID: ' uid = UidString '\n')?
  fields *= SDocNodeField
  (
    'RELATIONS:' '\n'
    (relations += Reference)
  )?
;
```

But this would have a similar surprising effect as MID has right know:

[...]

PR #1903, opened 2024-06-30
backend/sdoc: SDocNode: Make MID a normal SDocField

stanislaw:
Follow-up to https://github.com/strictdoc-project/strictdoc/issues/1897#issuecomment-2187562062


stanislaw:
cc @haxtibal

PR #1904, opened 2024-06-30
Add parentheses to UID regex

haxtibal:
Follow up to #1902.


haxtibal:
Just noticed I broke my own document with #1902. It roughly goes like
```
[REQUIREMENT]
UID: CR 1.1
STATEMENT: A statement.

[REQUIREMENT]
UID: CR 1.1 Enhancement (1)
STATEMENT: One more thing.
```

Can we allow parentheses? Do you know more characters that should be allowed right now?

PR #1908, opened 2024-07-01
rst: Fix link followed by non-ws/punctuation

### LINKED ISSUE #1907: LINKs: Unexpected output / crash if appending "s" to [LINK:] for plural
Sometimes we want to use an anchor as plural by appending "s", as in `[LINK: ROLE-1]s`. The example document below creates bad HTML that contains literal `:rawhtml:` when viewed in a browser, like ``🔗 System Integrator`s are expected to provide FIDO2 compliant hardware tokens. See :rawhtml:`🔗 Roles.``

A modified version of the example, where one removes the final `[LINK: ROLES]`, doesn't export at all but crashes with
```
error: problems when converting RST to HTML:
RST markup syntax error on line 1: Inline interpreted text or phrase reference start-string without end-string.
```

Example document to reproduce:
```
[DOCUMENT]
TITLE: Unexpected HTML for plural links

[SECTION]
UID: ROLES
TITLE: Roles

[TEXT]
STATEMENT: >>>
[ANCHOR: ROLE-1, System Integrator]
Service provider that specializes in bringing together component subsystems into a whole.
<<<

[/SECTION]

[SECTION]
TITLE: Assumptions

[TEXT]
STATEMENT: >>>
[LINK: ROLE-1]s are expected to provide FIDO2 compliant hardware tokens.
See [LINK: ROLES].
<<<

[/SECTION]
```

haxtibal:
When rendering rst fragments to HTML, InlineLinks are first converted to ``':rawhtml:`<a href="#TGT">link text</a>`'``, then are rendered to HTML by docutils `publish_parts`. That rawhtml construct is a special form of inline markup, for which rst defines recognition [rules](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#inline-markup-recognition-rules). Rule 7 says

> Inline markup end-strings must end a text block or be immediately followed by whitespace, one of the ASCII characters - . , : ; ! ? \ / ' " ) ] } > or a similar non-ASCII punctuation character. [20]

That's why LINK followed by character produces error or odd HTML. To get around, we can use rst's [escaping mechanism](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#escaping-mechanism).

> Escaped non-white characters are prevented from playing a role in any markup interpretation. The escaped character represents the character itself.

The code change detects if an inline links is followed by non-whitespace or non-punctuation character and adds in a backslash if so.

Punctuation characters in the non-latin range are not explicitly checked and thus also joined with "\". This is unnecessary from RST syntax point of view, but doesn't hurt as it just maps them to themselves.

Fixes #1907.

### DIFF
--- a/strictdoc/export/html/renderers/markup_renderer.py
+++ b/strictdoc/export/html/renderers/markup_renderer.py
@@ -20,6 +20,8 @@ from strictdoc.export.rst.rst_to_html_fragment_writer import (
     RstToHtmlFragmentWriter,
 )
 
+VALID_AFTER_INLINE_MARKUP = " \t-.,:;!?\\/'\")]}>"
+
 
 class MarkupRenderer:
     @staticmethod
@@ -109,10 +111,17 @@ class MarkupRenderer:
         if (document_type, node_field, truncated) in self.cache:
             return self.cache[(document_type, node_field, truncated)]
 
+        prev_part = None
         parts_output = ""
         for part in node_field.parts:
             if isinstance(part, str):
-                parts_output += part
+                if (
+                    isinstance(prev_part, InlineLink)
+                    and part[0] not in VALID_AFTER_INLINE_MARKUP
+                ):
+                    parts_output += f"\\{part}"
+                else:
+                    parts_output += part
             elif isinstance(part, InlineLink):
                 linkable_node = (
                     self.traceability_index.get_linkable_node_by_uid(part.link)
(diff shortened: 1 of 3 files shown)


haxtibal:
@stanislaw I'd add a test... but where? Can you make a suggestion?

stanislaw:
> @stanislaw I'd add a test... but where? Can you make a suggestion?

The best group of integration tests is here: `rst_markup_to_html`.

haxtibal:
> The best group of integration tests is here: `rst_markup_to_html`.

Thanks, added it there.

In the just pushed version I notably extended the allowed character list to the whole latin part of rule 7. For example, the first version would have inserted a "\" in between link followed by ":". This was valid too (maps : to itself), but I think it's better to limit places where we insert "\" as much as easily detectable.

PR #1909, opened 2024-07-04
rst: Render links and anchors in node fields

### LINKED ISSUE #1888: RST export: LINKs from requirements and other nodes: Resolve and export them correctly

haxtibal:
This worked only for freetext. Now it works for any node including [TEXT] and [REQUIREMENT].

The change includes some drive-by fixes:
- rst output now renders links immediately followed by character (same story as #1907).
- Escaping text following inline markup missed a check for newline, which is now added and effective for rst and HTML.

Fixes #1888.


haxtibal:
> rst output now includes custom multiline fields after comment fields.

This claim is wrong. It was there before, but did not and still doesn't render links. Will fix that, please wait.

stanislaw:
Thanks!

PR #1910, opened 2024-07-06
views: Support custom views for passthrough

haxtibal:
This enhances #1509. It makes the custom views option available for the passthrough command. Command line syntax is the same as for export:
```
strictdoc passthrough --view MY_VIEW .
```

We can now do an sdoc -> sdoc transformation where only fields from a given custom view are retained in the output sdoc.

### DIFF
--- a/strictdoc/backend/sdoc/writer.py
+++ b/strictdoc/backend/sdoc/writer.py
@@ -9,7 +9,10 @@ from strictdoc.backend.sdoc.models.document import SDocDocument
 from strictdoc.backend.sdoc.models.document_config import DocumentConfig
 from strictdoc.backend.sdoc.models.document_from_file import DocumentFromFile
 from strictdoc.backend.sdoc.models.document_grammar import DocumentGrammar
-from strictdoc.backend.sdoc.models.document_view import DefaultViewElement
+from strictdoc.backend.sdoc.models.document_view import (
+    DefaultViewElement,
+    ViewElement,
+)
 from strictdoc.backend.sdoc.models.inline_link import InlineLink
 from strictdoc.backend.sdoc.models.node import (
     SDocCompositeNode,
@@ -33,6 +36,7 @@ from strictdoc.backend.sdoc.models.type_system import (
 )
 from strictdoc.core.document_iterator import DocumentCachingIterator
 from strictdoc.core.document_meta import DocumentMeta
+from strictdoc.core.project_config import ProjectConfig
 from strictdoc.helpers.cast import assert_cast
 from strictdoc.helpers.string import ensure_newline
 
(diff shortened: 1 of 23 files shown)


haxtibal:
WIP because I have to check for completeness yet. You can comment right now if you see problems with the approach.

The few interesting design decisions:
- Command line takes one `--view`, not many `--views`, for consistency with the export command.
- Filtering happens in `SDWriter`, as opposed to `TraceabilityIndexBuilder` as --filter-requirements does. That's because a view shall only affect the output, not the inner model.
- We need to propagate project config. `SDWriter` gets it per constructor injection. It's a boring change but causes the bulk of the diff.

stanislaw:
> WIP because I have to check for completeness yet. You can comment right now if you see problems with the approach.

The approach is good. The price of introducing project_config to the SDWriter *feels* a little to high but doing without, by only passing a view= parameter that would have to be added to all the writer's functions, is not better.

The only further work we should do is to move the `filter-requirements/sections` and `view` to the export command where the whole export would benefit from these options. The passthrough command would only stay with the `--free-text-to-text` which is an exception for migrating for FREETEXT to TEXT.

> The few interesting design decisions:
> 
>     * Command line takes one `--view`, not many `--views`, for consistency with the export command.

This is fine but I think we should not keep it with `passthrough` unless you have good arguments for keeping it.

>     * Filtering happens in `SDWriter`, as opposed to `TraceabilityIndexBuilder` as --filter-requirements does. That's because a view shall only affect the output, not the inner model.

This makes sense.

>     * We need to propagate project config. `SDWriter` gets it per constructor injection. It's a boring change but causes the bulk of the diff.

It is ok. Happens all the time with similar changes.

I am merging this change as-is but let's discuss if we want to clean up the `passthrough` from `--filter-requirements`, `--filter-sections` and `--view` and have them working in `export`.

haxtibal:
> I am merging this change as-is

Ok, thanks!

> but let's discuss if we want to clean up the `passthrough` from `--filter-requirements`, `--filter-sections` and `--view` and have them working in `export`.

Your reasoning on Discord for it was good. From what I've seen it'll be another mostly boring change with bigger diff. Would be good to have it prior to Q3. I'll try to help.

stanislaw:
> > I am merging this change as-is
> 
> Ok, thanks!
> 
> > but let's discuss if we want to clean up the `passthrough` from `--filter-requirements`, `--filter-sections` and `--view` and have them working in `export`.
> 
> Your reasoning on Discord for it was good. From what I've seen it'll be another mostly boring change with bigger diff. Would be good to have it prior to Q3. I'll try to help.

Thanks! Many of the open tickets are mostly boring changes :) Here's a ticket to track this: https://github.com/strictdoc-project/strictdoc/issues/1913.

PR #1917, opened 2024-07-10
links: Consistent anchor id whitespace replacement

### LINKED ISSUE #1916: LINKs: Anchors with IDs containing whitespace are not referrable in exported HTML
This sdoc

```
[DOCUMENT]
TITLE: Link Test

[TEXT]
STATEMENT: >>>
[ANCHOR: ID 1]
<<<

[REQUIREMENT]
UID: REQ-1
STATEMENT: >>>
[LINK: ID 1]
<<<
```

generates an anchor like `<sdoc-anchor id="ID 1" node-role="section" data-uid="ID 1" data-anchor="ID 1" style="top:unset">...</sdoc-anchor>` and refers to it with `<a href="#ID-1">🔗&nbsp;ID 1</a>`.

Note how `id` contains a whitespace, but `href` has the whitespace replaced with dash. Replacing with dash makes sense because `UID` may contain whitespace in sdoc but `id` [must not contain whitespaces](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id) in HTML. We just have to do the replacement on both ends.

Will try to fix that.

haxtibal:
Whitespaces are valid in StrictDoc UIDs, but are not valid in HTML element ids. Therefore we have to sanitize them during HTML export. This was already done in most places, but not for anchor IDs in rst fragments. In consequence referencing links couldn't find their target.

This adds whitespace to dash replacement for rst-for-html anchors. It reuses LinkRenderer for consistent results.

Fixes #1916.

### DIFF
--- a/strictdoc/export/html/templates/rst/anchor.jinja
+++ b/strictdoc/export/html/templates/rst/anchor.jinja
@@ -5,7 +5,8 @@
 
 .. raw:: html
 
-    <sdoc-anchor id="{{ anchor.value }}" node-role="section" data-uid="{{ anchor.value }}" data-anchor="{{ anchor.value }}" style="top:unset">
+    {% set local_anchor = link_renderer.render_local_anchor(anchor) -%}
+    <sdoc-anchor id="{{ local_anchor }}" node-role="section" data-uid="{{ local_anchor }}" data-anchor="{{ local_anchor }}" style="top:unset">
     {%- set incoming_links = traceability_index.get_incoming_links(anchor) -%}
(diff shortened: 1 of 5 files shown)


haxtibal:
I'm not yet sure if it's correct to set the replaced value not only to `id`, but also to `data-uid` and `data-anchor`.  Will have to check how these data attributes are used.

stanislaw:
It should be a safe step forward. Using the opportunity, I added a small e2e test to exercise this behavior. Thanks for contributing!

Issue #1920, opened 2024-07-16
Epic: Correct escaping in HTML output


haxtibal:
- [x] https://github.com/strictdoc-project/strictdoc/pull/1929
- [x] https://github.com/strictdoc-project/strictdoc/pull/1921
- [x] https://github.com/strictdoc-project/strictdoc/pull/1936
- [x] HTML escaping: Project tree. https://github.com/strictdoc-project/strictdoc/pull/1943
- [x] HTML escaping: add itest: Traceability Matrix screen. https://github.com/strictdoc-project/strictdoc/pull/1943
- [x] HTML escaping: add itest: Search screen. https://github.com/strictdoc-project/strictdoc/pull/1954
- [x] HTML escaping: add itest: Statistics screen. https://github.com/strictdoc-project/strictdoc/pull/1954
- [x] HTML escaping: add itest: Diff/Changelog screens. https://github.com/strictdoc-project/strictdoc/pull/1993

----

To reproduce, generate an html export of the following document
```
[DOCUMENT]
TITLE: Written <b>bold</b>

[SECTION]
TITLE: Some text is <invisible>

[REQUIREMENT]
UID: REQ-1
TITLE: <script>alert(1)</script>
STATEMENT: >>>
We can also execute scripts.
<<<

[/SECTION]
```

I assume this is a bug, not a feature...

`html.escape` is already used in some places. I'll have a look and try to catch more cases.

stanislaw:
Thanks for reporting this. It is a lack of feature 😄 

The html.escape was implemented for multiline fields but not for single-line. Let's implement it for single-line as well.

haxtibal:
> The html.escape was implemented for multiline fields

IIUC multiline was implemented for the UI forms in !1018. Multiline in static views (like components/requirement/index.jinja) is escaped too, but for a different reason: We pipe it through docutils, which does escaping under the hood.

I thought about when and where is the best place to do the escaping:
- `backend.sdoc` (`SDocNodeField.enumerate_all_fields_escaped` exists right now) is central, but IMO it's wrong because it adds HTML specific stuff to a core component
- call `html.escape` in Jinja2 templates where needed (lots of places)
- pipe through [|e](https://jinja.palletsprojects.com/en/3.0.x/templates/#working-with-manual-escaping) in Jinja2 templates (same lot of places, but keeps code cleaner)
[...]

PR #1921, opened 2024-07-18
Use Jinja2 autoescaping 

haxtibal:
This enables Jinja2 autoescaping everywhere, except for the diff screen which relies on pre generated HTML. The MR is not meant to be merged as-is.

It fixes the issues observed in #1920. Most integration tests pass. Tests fail for UI forms, presumably because they call `html.escape` manually and now are escaped twice.

Jinja2 [recommneds autoescaping](https://jinja.palletsprojects.com/en/3.1.x/api/#basics):

> In future versions of Jinja we might enable autoescaping by default for security reasons.
> As such you are encouraged to explicitly configure autoescaping now instead of relying on the default.

Templates access way more functions and variables than just `enumerate_all_fields()`. All need to be escaped. This does it automatically without risk of forgetting some places.

This MR changes some formerly explicit escaping:
- `enumerate_all_fields_escaped` was the `html.escape()` variant of `enumerate_all_fields`. It's only evaluated in Jinja2 templates, therefore completely covered by autoescaping now. This also eliminates the need for `get_text_value_escaped`, which was only used by `enumerate_all_fields_escaped`.
- `RequirementFormField.field_escaped_value` was the `html.escape()` variant of `field_unescaped_value`. `field_escaped_value` is only evaluated in Jinja2 templates, therefore completely covered by autoescaping now.
- `SearchScreenViewObject.search_value` was explictly `html.escape()`ed. `search_value` is only read by Jinja2 templates, therefore completely covered by autoescaping now.
- `SourceFileViewHTMLGenerator` changes some pygments output lines. pygements output is escaped, but we still have to escape our own changes. Changed `html.escape` to `markupsafe.escape` because that's what Jinja2 uses. It's better to use the same escaping function everywhere.
- `TextToHtmlWriter` escaped plain text from multiline statements. We still need to do that. Changed `html.escape` to `markupsafe.escape` for same reason as above.

Considerations about what to escape
- `render_*_link` make strings used as URL in href="`xyz`" . It's derived from filenames and UIDs. UIDs are restricted by a grammar regex. But Unix files can have weird names like `document"<injected>.sdoc`. So generated links should be URL encoded. HTML encoding is then not strictly necessary (wouldn't hurt either; could be considered as safety net). Suggestion: Do proper URL encoding in another PR; for now autoescaping prevents at least injection attacks.

TODO:
- [x] investigate failed ruff checks: Unrelated, new or changed ruff rule PLR1714
- [x] investigate failed integration tests (works locally): unrelated, we need either old filecheck 0.0.24 or fixed filecheck > 1.0.0
- [ ] investigate failed end2end tests
- [ ] try to replace some `| safe` with `Markup()` in code
- [ ] reconsider if autoescaping could be used for the diff screen
- [ ] Templates often render a variable that contains the output of calling a nested `Template.render()` programmatically. This output can be considered safe, yet we have to wrap it in `Markup`. Is there a way to automatically flag `Template.render()` as safe? Should already be that way: "Jinja functions (macros, super, self.BLOCKNAME) always return template data that is marked as safe."

### DIFF
--- a/docs/strictdoc_25_design.sdoc
+++ b/docs/strictdoc_25_design.sdoc
@@ -136,3 +136,39 @@ One important implementation detail of Arpeggio that influences StrictDoc user e
 [/FREETEXT]
 
 [/SECTION]
+
+[SECTION]
+TITLE: HTML Escaping
+
+[TEXT]
+STATEMENT: >>>
+StrictDoc uses Jinja2 autoescaping_ for HTML output. `Template.render`_ calls
+will escape any Python object unless it's explicitly marked as safe.
+
+Good to know for a start:
+
+- If a Python object intentionally contains HTML it must be marked as safe
+  to bypass autoescaping. Templates can do this by piping to safe_, or Python code
+  can do it by wrapping an object into `markupsafe.Markup`_.
+- Passing text to the `Markup() <markupsafe.Markup_>`_ constructor marks that text
+  as safe, but *does not escape* it.
+- Text can be explicitly escaped with `markupsafe.escape`_. It's similar to
+  `html.escape`_, but the result is immediately marked safe.
+- `markupsafe.Markup`_ is responsible for some "magic". It's a :code:`str` subclass
+  with the same methods, but escaping arguments. For example,
+  :code:`"> " + Markup("<div>safe</div>")` will turn into :code:`"&gt; <div>safe</div>"`,
+  thanks to :code:`__radd__` in this specific case. To prevent escaping,
+  you would use :code:`Markup("> ") + Markup("<div>safe</div>")`. Basically the
+  same magic happens in templates when using safe_.
+- See also `Working with Automatic Escaping`_.
+
+.. _autoescaping: https://jinja.palletsprojects.com/en/latest/api/#autoescaping
+.. _Working with Automatic Escaping: https://jinja.palletsprojects.com/en/latest/templates/#working-with-automatic-escaping
+.. _markupsafe.Markup: https://markupsafe.palletsprojects.com/en/latest/escaping/#markupsafe.Markup
+.. _markupsafe.escape: https://markupsafe.palletsprojects.com/en/latest/escaping/#markupsafe.escape
+.. _safe: https://jinja.palletsprojects.com/en/latest/templates/#jinja-filters.safe
+.. _Template.render: https://jinja.palletsprojects.com/en/latest/api/#jinja2.Template.render
+.. _html.escape: https://docs.python.org/3/library/html.html#html.escape
+<<<
+
+[/SECTION]
--- a/strictdoc/export/html/generators/view_objects/document_screen_view_object.py
+++ b/strictdoc/export/html/generators/view_objects/document_screen_view_object.py
@@ -3,7 +3,8 @@ from dataclasses import dataclass
 from datetime import datetime
 from typing import List, Optional, Union
 
-from jinja2 import Environment, Template
+from jinja2 import Template
+from markupsafe import Markup
 
 from strictdoc import __version__
 from strictdoc.backend.sdoc.models.document import SDocDocument
@@ -15,6 +16,7 @@ from strictdoc.core.file_tree import Folder
 from strictdoc.core.project_config import ProjectConfig
 from strictdoc.core.traceability_index import TraceabilityIndex
 from strictdoc.export.html.document_type import DocumentType
+from strictdoc.export.html.html_templates import JinjaEnvironment
 from strictdoc.export.html.renderers.link_renderer import LinkRenderer
 from strictdoc.export.html.renderers.markup_renderer import MarkupRenderer
 from strictdoc.server.helpers.turbo import render_turbo_stream
@@ -63,68 +65,71 @@ class DocumentScreenViewObject:
     def has_included_document(self):
         return len(self.document.included_documents) > 0
 
-    def render_screen(self, jinja_environment: Environment):
+    def render_screen(self, jinja_environment: JinjaEnvironment) -> Markup:
         if self.document_type.is_document:
             if self.document.config.layout == "Website":
-                template = jinja_environment.get_template(
-                    "website/document/index.jinja"
+                return jinja_environment.render_template_as_markup(
+                    "website/document/index.jinja", view_object=self
                 )
-            else:
-                template = jinja_environment.get_template(
-                    "screens/document/document/index.jinja"
-                )
+            return jinja_environment.render_template_as_markup(
+                "screens/document/document/index.jinja", view_object=self
+            )
         elif self.document_type.is_table():
-            template = jinja_environment.get_template(
-                "screens/document/table/index.jinja"
+            return jinja_environment.render_template_as_markup(
+                "screens/document/table/index.jinja", view_object=self
             )
         elif self.document_type.is_trace():
-            template = jinja_environment.get_template(
-                "screens/document/traceability/index.jinja"
+            return jinja_environment.render_template_as_markup(
+                "screens/document/traceability/index.jinja", view_object=self
             )
         elif self.document_type.is_deeptrace:
-            template = jinja_environment.get_template(
-                "screens/document/traceability_deep/index.jinja"
+            return jinja_environment.render_template_as_markup(
+                "screens/document/traceability_deep/index.jinja",
+                view_object=self,
             )
         elif self.document_type.is_pdf():
-            template = jinja_environment.get_template(
-                "screens/document/pdf/index.jinja"
+            return jinja_environment.render_template_as_markup(
+                "screens/document/pdf/index.jinja", view_object=self
             )
         else:
             raise NotImplementedError(self.document_type)
-        return template.render(view_object=self)
 
-    def render_table_screen(self, jinja_environment: Environment):
-        template = jinja_environment.get_template(
-            "screens/document/table/index.jinja"
+    def render_table_screen(
+        self, jinja_environment: JinjaEnvironment
+    ) -> Markup:
+        return jinja_environment.render_template_as_markup(
+            "screens/document/table/index.jinja", view_object=self
         )
-        return template.render(view_object=self)
 
-    def render_trace_screen(self, jinja_environment: Environment):
-        template = jinja_environment.get_template(
-            "screens/document/traceability/index.jinja"
+    def render_trace_screen(
+        self, jinja_environment: JinjaEnvironment
+    ) -> Markup:
+        return jinja_environment.render_template_as_markup(
+            "screens/document/traceability/index.jinja", view_object=self
         )
-        return template.render(view_object=self)
 
-    def render_updated_screen(self, jinja_environment: Environment) -> str:
-        template = jinja_environment.get_template(
+    def render_updated_screen(
+        self, jinja_environment: JinjaEnvironment
+    ) -> Markup:
+        output = jinja_environment.render_template_as_markup(
             "actions/"
             "document/"
             "create_requirement/"
-            "stream_created_requirement.jinja.html"
+            "stream_created_requirement.jinja.html",
+            view_object=self,
         )
-        output = template.render(view_object=self)
 
-        toc_template = jinja_environment.get_template(
-            "actions/document/_shared/stream_updated_toc.jinja.html"
+        output += jinja_environment.render_template_as_markup(
+            "actions/document/_shared/stream_updated_toc.jinja.html",
+            view_object=self,
         )
-        output += toc_template.render(view_object=self)
 
         return output
 
     def render_updated_nodes_and_toc(
         self,
         nodes: List[Union[SDocDocument, SDocNode]],
-        jinja_environment: Environment,
+        jinja_environment: JinjaEnvironment,
     ) -> str:
         output: str = ""
 
@@ -139,20 +144,22 @@ class DocumentScreenViewObject:
                     template_folder = "requirement"
             else:
                 raise NotImplementedError
-            template = jinja_environment.get_template(
-                f"components/{template_folder}/index_extends_node.jinja"
+            content = jinja_environment.render_template_as_markup(
+                f"components/{template_folder}/index_extends_node.jinja",
+                view_object=self,
+                node=node_,
             )
             output += render_turbo_stream(
-                content=template.render(view_object=self, node=node_),
+                content=content,
                 action="replace",
                 target=f"article-{node_.reserved_mid}",
             )
 
-        toc_template = jinja_environment.get_template(
-            "screens/document/_shared/toc.jinja"
+        toc_content = jinja_environment.render_template_as_markup(
+            "screens/document/_shared/toc.jinja", view_object=self
         )
         output += render_turbo_stream(
-            content=toc_template.render(view_object=self),
+            content=toc_content,
             action="update",
             target="frame-toc",
         )
@@ -160,24 +167,24 @@ class DocumentScreenViewObject:
         return output
 
     def render_update_document_content_with_moved_node(
-        self, jinja_environment: Environment, moved_node
-    ) -> str:
-        template = jinja_environment.get_template(
-            "screens/document/document/frame_document_content.jinja.html"
+        self, jinja_environment: JinjaEnvironment, moved_node
+    ) -> Markup:
+        content = jinja_environment.render_template_as_markup(
+            "screens/document/document/frame_document_content.jinja.html",
+            view_object=self,
         )
         output = render_turbo_stream(
-            content=template.render(view_object=self),
+            content=content,
             action="replace",
             target="frame_document_content",
         )
-        toc_template = jinja_environment.get_template(
-            "actions/document/_shared/stream_updated_toc.jinja.html"
+        toc_content = jinja_environment.render_template_as_markup(
+            "actions/document/_shared/stream_updated_toc.jinja.html",
+            view_object=self,
+            last_moved_node_id=moved_node.reserved_mid,
         )
         output += render_turbo_stream(
-            toc_template.render(
-                view_object=self,
-                last_moved_node_id=moved_node.reserved_mid,
-            ),
+            toc_content,
             action="update",
             target="frame-toc",
         )
@@ -205,10 +212,10 @@ class DocumentScreenViewObject:
                 return True
         return False
 
-    def render_url(self, url: str):
-        return self.link_renderer.render_url(url)
+    def render_url(self, url: str) -> Markup:
+        return Markup(self.link_renderer.render_url(url))
 
-    def render_node_link(self, node, context_document, document_type):
+    def render_node_link(self, node, context_document, document_type) -> str:
         assert node is not None, node
         return self.link_renderer.render_node_link(
             node, context_document, document_type
@@ -219,7 +226,7 @@ class DocumentScreenViewObject:
         document: SDocDocument,
         context_document: SDocDocument,
         document_type_string: str,
-    ):
+    ) -> str:
         assert document is not None, document
         return self.link_renderer.render_node_link(
             document, context_document, DocumentType(document_type_string)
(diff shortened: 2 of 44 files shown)


stanislaw (commented):
This looks reasonable. Please proceed with the Jinja2 autoescape approach 👍

haxtibal:
Integration tests fail because the recently released filecheck 1.0.0 [doesn't support `--dump-input`](https://github.com/AntonLydike/filecheck/issues/19). This is a regression unrelated to this PR.

stanislaw:
> Integration tests fail because the recently released filecheck 1.0.0 https://github.com/AntonLydike/filecheck/issues/19. This is a regression unrelated to this PR.

Sorry for the inconvenience.

We are in the process of transferring the filecheck project to a group of developers who want to actively maintain and develop their own version of `filecheck` instead of `FileCheck.py` that was previously developed by me. We were too quick to release a 1.0.0, and it doesn't have the dump-input implemented yet. 

I have reported this here: https://github.com/AntonLydike/filecheck/issues/19#issuecomment-2241237118.

strictdoc/core/document_iterator.py
```diff
@@ -9,6 +9,19 @@
 )
 from strictdoc.backend.sdoc.models.section import FreeText, SDocSection
 
+DocumentAny = NewType(
```
stanislaw, line 12: This will probably not work. Let me handle this myself in a separate PR.

strictdoc/core/file_tree.py
```diff
@@ -194,9 +194,7 @@ def find_files_with_extensions(
                 if (
                     not d.startswith(".")
                     and (d == "_static" or not d.startswith("_"))
-                    and d != "build"
-                    and d != "output"
-                    and d != "Output"
+                    and d not in ("build", "output", "Output")
```
stanislaw, line 197: Let's separate the trivial linting fixes from the actual meat of this PR. I opened a separate PR to first sort out the new Ruff checks: https://github.com/strictdoc-project/strictdoc/pull/1922.

strictdoc/core/document_iterator.py
haxtibal: Yes, fixed it already locally. Needs to be alias, not newtype. It's an unrelated drive by change anyways. Thanks if you handle it separately.

strictdoc/core/file_tree.py
haxtibal: Thanks as well

strictdoc/core/document_iterator.py
stanislaw: Yes, I also see that you don't really need it for this PR. The change is nice though, so I will handle it.

haxtibal:
The latest commit "Mark all template.render() output as Markup" is to be discussed: We have to flag every programmatically generated HTML string as "safe", otherwise autoescaping would wrongly process them. This notably applies to all `template.render()` calls. I hoped Jinja2 would automatically mark the render output as safe, but seemingly not. So I tried to centralize Markup wrapping a bit. The result is a systematic but big diff. Would it be ok for you?

I asked Jinja2 folks whether such a feature could be implemented, see pallets/jinja#2003.

strictdoc/export/html/html_templates.py
```diff
@@ -17,13 +17,31 @@
...
+class JinjaEnvironment:
+    environment: Environment
+
+    def __init__(self, environment: Environment):
+        self.environment = environment
+
+    def get_template(self, *args, **kwargs):
+        return self.environment.get_template(*args, **kwargs)
+
+    def render_template_as_markup(
```
stanislaw, line 37: I think we are very fine with adding this method to control what is passed to or taken from the render() call.

One thing that I am slightly missing to understand:

Does it mean that this solution replaces the need to do `autoescape=True` because you anyway wrap the output into Markup?

Or the `autoescape=True` is still needed?

strictdoc/backend/reqif/p01_sdoc/sdoc_to_reqif_converter.py
```diff
@@ -235,21 +235,22 @@ def convert_document_tree(
             current_hierarchy = root_hierarchy
 
             # FIXME: ReqIF must export complete documents including fragments.
-            for node in document_iterator.all_content(
+            for node_ in document_iterator.all_content(
```
stanislaw, line 238: Now you can rebase.

strictdoc/server/helpers/turbo.py
```diff
@@ -9,4 +12,4 @@ def render_turbo_stream(content: str, action: str, target: str):
   </template>
 </turbo-stream>
 """
-    return turbo_stream
+    return Markup(turbo_stream)
```
stanislaw, line 15: Related to my other comment's question: Does this mean that the Markup will be applied two times: one time here and one time if the `content` is rendered with `render_template_as_markup`?

stanislaw:
> The latest commit "Mark all template.render() output as Markup" is to be discussed: We have to flag every programmatically generated HTML string as "safe", otherwise autoescaping would wrongly process them. This notably applies to all `template.render()` calls. I hoped Jinja2 would automatically mark the render output as safe, but seemingly not. So I tried to centralize Markup wrapping a bit. The result is a systematic but big diff. Would it be ok for you?

I like the idea of `render_template_as_markup` because it also reduces the boilerplate everywhere.

I commented two times, asking about the possible redundancy of Markup / autoescape=True. I would like you to confirm that both mechanisms have to be used by us (Markup and autospace=True).

> I asked Jinja2 folks whether such a feature could be implemented, see [pallets/jinja#2003](https://github.com/pallets/jinja/issues/2003).

I could not understand what you are referring to in your example there:

> The Template knows inner is created with autoescape=True. Couldn't it mark the result of render automatically as safe?

How can it make the result of render automatically if the result is a simple `str` object?

```py
>>> inner = Template('<b>{{ first_name }}</b> {{ last_name }}', autoescape=True).render(first_name='John', last_name="Doe")
>>> print(type(inner))
<class 'str'>
```

----

While reviewing your PR, I considered an alternative approach that could be possible instead of doing the escaping with Jinja.

I recently encountered several implicit conversions with Jinja, when an object is implicitly called `__str__` and so something incorrect is rendered to Jinja without me or the integration tests noticing this magic stuff.

An alternative could be that the whole StrictDoc uses a wrapper of a `str` class, e.g. `sdoc_str` that would disable the `__str__` method with an explicit assert. Instead the safe version of a string would be implemented as `Markup(self)` and only this method would be called in each and every Jinja template.

With this approach, the diff could be even bigger but in return we would get a very good control over the strings that we really want to use everywhere. It is in the same direction like the code of `field_value`/`field_unescaped_value` that you are removing now but this escaping stuff would be inside the `sdoc_str`.

I have confidence that this approach could work just fine but I don't have enough data to compare the advantages/disadvantages of using Jinja's escaping magic vs packing that explicitly in `sdoc_str`.

The `sdoc_str` approach would fit the current philosophy of SDoc of doing as many things explicit as possible. The Jinja autoescaping stuff is a different approach but also a capable one.

strictdoc/export/html/html_templates.py
haxtibal: Short comment only as I'm afk. Yes autoescape ist needed. It works like this:

1. autoescape escapes every object that comes into the template from the Python side
2. except if the object is of type markupsafe.Markup, which tells autoescape "Don't touch me"

The Markup type is a very str compatible wrapper around str. isinstance(Markup("Test"), str) is True. Wrapping a str into Markup does NOT escape the str. For more details see the markupsafe project.

Does this address your question?

stanislaw: Your answer addresses my question to 80% but the remaining 20% is something that I see contradicting, given the experiment below.

To understand the behavior of Jinja and markupsafe I went to experiment with you branch a little bit and found something that I don't understand given your answer.

See this commit and the comment that I made:

```
{# WIP: DO NOT UNDERSTAND WHY adding |safe here makes the section title to be escaped #}
```

https://github.com/strictdoc-project/strictdoc/commit/66a380bdca13bd699969d1c8c6297b41802e5036

I cannot understand why the new itest passes when the `|safe` is there, i.e., the section does get escaped, and the other way round - if I remove `|safe`, it is NOT escaped.

My understanding is that `safe` means exactly the opposite of `do escape` because it says `do not escape me, I am escaped already`.

What am I missing?

strictdoc/export/html/renderers/markup_renderer.py
```diff
@@ -141,7 +143,7 @@ def render_node_field(
         output = self.fragment_writer.write(parts_output)
         self.cache[(document_type, node_field, truncated)] = output
 
-        return output
+        return Markup(output)
```
stanislaw, line 146: Let's add a comment here that explains that docutils escapes stuff by itself.

stanislaw: Let's add a type hint to the .write() methods that it turns Markup, not str to make it explicit in the type system that the stuff is escaped already?

strictdoc/export/html/html_templates.py
haxtibal: I think the key to understand it is [another property](https://markupsafe.palletsprojects.com/en/latest/escaping/#markupsafe.Markup) of the `Markup` type.

> This [the Markup type] is a subclass of [str](https://docs.python.org/3/library/stdtypes.html#str). It has the same methods, *but escapes their arguments*

Consider this minimal code:

```python
from markupsafe import Markup

Markup("Markup's str functions (e.g. __add__) consider raw str args as unsafe and escape them: ") + "<div>"
Markup("To prevent it, mark args as safe: ") + Markup("<div>")
```

In your example, you set `autoescape=False`. So expectation is that title would not be escaped. But when we apply the `| safe` filter to `".&nbsp;"`, we create a `Markup(".&nbsp;")` object. If we add the title (a raw `str`) to it, `Markup.__add__` will escape `title` during the `+` operation.

Does this make sense?

stanislaw: Now it makes complete sense. I just tried it this way, and this is even closer to the line in my commit:

```
>>> "<div>" + Markup("Markup's str functions (e.g. __add__) consider raw str args as unsafe and escape them: ")
Markup("&lt;div&gt;Markup's str functions (e.g. __add__) consider raw str args as unsafe and escape them: ")
```

I kind of felt yesterday that it should work this way but didn't go to check it myself, having asked my question above instead. Thanks for explaining!

----

Could you also explain how can it work in both directions? I just tried this naive thing and it is not working the way Markup does above.

```py
class GFG(str):

    def __init__(self, val):
        self.val = val

    def __add__(self, val2):
        return GFG(self.val + val2.val)


obj1 = "GEEKS"
obj2 = GFG("ForGeeks")
obj3 = obj1 + obj2
print(obj3.val)
```

```
Traceback (most recent call last):
  File "/home/stanislav/sandbox/volatile/pythonProject2/main.py", line 13, in <module>
    print(obj3.val)
AttributeError: 'str' object has no attribute 'val'
```

----


I understand the behavior now perfectly, and I am now wondering if there is a way to improve the readability of all affected lines such that one does not have to mentally do this rule calculation. I.e.,

```
{%- set field_content_ = field_content_ + sdoc_entity.context.title_number_string + ".&nbsp;"|safe -%}
```

Written this way, one really has to be trained into understanding that:

1) `|safe` only works on the immediate member, not the whole string.
2) there is `[another property](https://markupsafe.palletsprojects.com/en/latest/escaping/#markupsafe.Markup) of the Markup type` like you explained.

We are fine to go with all this written as-is in your PR. I just feel like these things are not explicit enough and somewhat increase the level of magic in the Jinja templates.

haxtibal: > Could you also explain how can it work in both directions?

Could you explain what you mean with "both directions"?

> I just tried this naive thing and it is not working the way Markup does above.

The `Markup` class doesn't aggregate a `str` attribute to hold the value. It rather *is* the value. This is accomplished by calling `str.__new__` from `Markup.__new__`. See [their implementation](https://github.com/pallets/markupsafe/blob/main/src/markupsafe/__init__.py#L81).

> |safe only works on the immediate member, not the whole string

What is the immediate member?

> I just feel like these things are not explicit enough and somewhat increase the level of magic in the Jinja templates.

I see your point. Yes it's a bit magic and I have no good idea how to make it more explicit without loosing the advantages of being less explicit. Besides adding comments and referring to documentation ([Autoescaping](https://jinja.palletsprojects.com/en/3.1.x/api/#autoescaping), [Working with Automatic Escaping](https://jinja.palletsprojects.com/en/3.1.x/templates/#working-with-automatic-escaping), [Working With Safe Text](https://markupsafe.palletsprojects.com/en/2.1.x/escaping/#markupsafe.Markup)). That could be considered lame though :)

stanislaw: > > Could you also explain how can it work in both directions?
> 
> Could you explain what you mean with "both directions"?

The order of Markup and str doesn't matter. This works:

```
>>> Markup("Markup's str functions (e.g. __add__) consider raw str args as unsafe and escape them: ") + "<div>"
Markup("Markup's str functions (e.g. __add__) consider raw str args as unsafe and escape them: &lt;div&gt;")
```

and this works:

```
>>> "<div>" + Markup("Markup's str functions (e.g. __add__) consider raw str args as unsafe and escape them: ")
Markup("&lt;div&gt;Markup's str functions (e.g. __add__) consider raw str args as unsafe and escape them: ")
```

My naive example (see my previous comment) doesn't allow the other direction, i.e., when the `str` goes first and it is added with Markup.

> > I just tried this naive thing and it is not working the way Markup does above.
> 
> The `Markup` class doesn't aggregate a `str` attribute to hold the value. It rather _is_ the value. This is accomplished by calling `str.__new__` from `Markup.__new__`. See [their implementation](https://github.com/pallets/markupsafe/blob/main/src/markupsafe/__init__.py#L81).
> 
> > |safe only works on the immediate member, not the whole string
> 
> What is the immediate member?
> 
> > I just feel like these things are not explicit enough and somewhat increase the level of magic in the Jinja templates.
> 
> I see your point. Yes it's a bit magic and I have no good idea how to make it more explicit without loosing the advantages of being less explicit. Besides adding comments and referring to documentation ([Autoescaping](https://jinja.palletsprojects.com/en/3.1.x/api/#autoescaping), [Working with Automatic Escaping](https://jinja.palletsprojects.com/en/3.1.x/templates/#working-with-automatic-escaping), [Working With Safe Text](https://markupsafe.palletsprojects.com/en/2.1.x/escaping/#markupsafe.Markup)). That could be considered lame though :)

Yes, please add the documentation, referencing these Jinja pages. I think a good place could be the Design document, not the User Guide, since this is mostly internal to StrictDoc. Add a section right after this one: https://strictdoc.readthedocs.io/en/stable/strictdoc_25_design.html#parsing-sdoc-files?

stanislaw: > What is the immediate member?

```
from jinja2 import Template

# Define a template string
template_string = """
{%- set field_content_ = foo2|safe + foo|safe -%}
{{ field_content_ }}
"""

# Create a template object
template = Template(template_string, autoescape=False)

# Render the template with the context
rendered_html = template.render(foo="FOO", foo2="<div>TEST</div>")

# Print the rendered HTML
print(rendered_html)
# <div>TEST</div>FOO
```

If you remove the first `safe`, then the escaped string is printed:

```
&lt;div&gt;TEST&lt;/div&gt;FOO
```

In your line, it is only the 
".&nbsp;" that is made to Markup, and that makes the whole expression to go escaped. This is exactly the behavior that is quite confusing when you have to read it right-to-left and do the mental magic. 

```
      {# WIP: DO NOT UNDERSTAND WHY adding |safe here makes the section title to be escaped #}
      {%- set field_content_ = field_content_ + sdoc_entity.context.title_number_string + ".&nbsp;"|safe -%}
```

haxtibal: >> Could you explain what you mean with "both directions"?
>
> The order of Markup and str doesn't matter.

IIUC it works by defining the [`__radd__`](https://docs.python.org/3/reference/datamodel.html#object.__radd__) method.

> These methods are called to implement the binary arithmetic operations (+, -, *, @, /, //, %, [divmod()](https://docs.python.org/3/library/functions.html#divmod), [pow()](https://docs.python.org/3/library/functions.html#pow), **, <<, >>, &, ^, |) with reflected (swapped) operands.

See [Markup.__radd__](https://github.com/pallets/markupsafe/blob/main/src/markupsafe/__init__.py#L139).

strictdoc/server/helpers/turbo.py
haxtibal: No, it's explained in [markupsafe.Markup](https://markupsafe.palletsprojects.com/en/latest/escaping/#markupsafe.Markup):

> Passing an object to the constructor converts it to text and wraps it to mark it safe *without escaping*.

strictdoc/export/html/html_templates.py
stanislaw: Yes, this is how it works! Super-clear now.

strictdoc/export/html/renderers/markup_renderer.py
stanislaw: Sounds good. I have been moving to 100% happy mypy when it works in `--strict` mode.

docs/strictdoc_25_design.sdoc
```diff
@@ -136,3 +136,39 @@ One important implementation detail of Arpeggio that influences StrictDoc user e
 [/FREETEXT]
 
 [/SECTION]
+
+[SECTION]
```
stanislaw, line 140: Perfect!

haxtibal:
> EDITED:
> 
> These are only thoughts. Let's move forward with Jinja's autoescape, safe, Markup. The alternative approach could work but let's keep it aside.

Sorry for not commenting on your alternative approach yet. It sounds interesting. I was quite focused here and wanted to get autoescaping in a workable state first. Will comment soon and suggest to hop back to #1920 for discussion to not interleave the autoescaping approach with the sdoc_str  approach too much.

strictdoc/export/html/generators/view_objects/document_screen_view_object.py
```diff
@@ -233,41 +240,43 @@ def render_standalone_document_link(
...
-        return "/".join((root_prefix, document_link))
+        return (
+            document_link
+            if len(root_prefix) == 0
+            else "/".join((root_prefix, document_link))
+        )
 
-    def render_static_url(self, url: str):
-        return self.link_renderer.render_static_url(url)
+    def render_static_url(self, url: str) -> str:
```
stanislaw, line 249: Why not `-> Markup` here?

stanislaw:
Thanks a lot for contributing this!

strictdoc/export/html/generators/view_objects/document_screen_view_object.py
haxtibal: It was an oversight. `-> str` is correct. Wrapping in `Markup()` is wrong. Reasoning is basically the same as in PR description, "Considerations about what to escape - render_*_link make...".

I intended to review and annotate all `strictdoc/export/html/generators/view_objects/*.py` for `str` vs. `Markup`, but was too slow. Another PR will follow.

Thanks for merging this early anyway. We use it for daily work now and experienced now failures so far.

stanislaw: Why would wrapping into `Markup` would be wrong in this case? Isn't it a string that doesn't have to be escaped?

haxtibal: > Isn't it a string that doesn't have to be escaped?

The string is used in places like `<link href="the_string">`. It usually contains the name of a file. There are two scenarios to consider:
- Non malicious: Let the filename contain a whitespace, e.g. `"some file.css"`. We end up with `<link href="some file.css">`, which is not a valid URL. The string must be URL escaped. `<link href="some%20file.css">` would be correct.
- Malicious: The filename is crafted so that causes injection. Let it be `'somefile.css"><script>bad()</script><div id="a` and we end up with `<link href="somefile.css"><script>bad()</script><div id="a">`.

One could argue the input string is trusted because it'll only ever contain hardcoded static file names and so the 2nd scenario can't happen. Then HTML escaping would not be necessary.

We should still consider URL escaping, which I suggested to do in a future MR due to current lack of time.

stanislaw: This makes sense. I didn't consider this escaping topic strongly before, and your work helps me understand this better and is moving the project to a very good direction.

It would be great if you could include this your comment to the design document as well. It helps a lot in understanding what and why we are doing with escaping. Also, this all contributes to at least the basic handling of the security topic by StrictDoc.

Issue #1912, opened 2024-07-07
2024-Q3/Q4: Code climate


stanislaw:
Previous: https://github.com/strictdoc-project/strictdoc/issues/1770
Next: https://github.com/strictdoc-project/strictdoc/issues/2068

---

- [x] Remove unused `render_free_text()`. https://github.com/strictdoc-project/strictdoc/pull/1924
- [x] https://github.com/strictdoc-project/strictdoc/pull/1927
- [x] https://github.com/strictdoc-project/strictdoc/pull/1955
- [x] `output/cache/` instead of `$TMPDIR`.
- [x] `edit_grammar_add_field_move_up_and_down_save`, https://github.com/strictdoc-project/strictdoc/actions/runs/10244497289/job/28337588222?pr=1932.
- [x] `edit_grammar_move_field_down`, https://github.com/strictdoc-project/strictdoc/actions/runs/10256651383/job/28376187324?pr=1932.

PR #1929, opened 2024-07-26
Fix html escaping of source line marks

haxtibal:
`SourceFileViewHTMLGenerator` escaping code had multiple issues:

- When marking each line from `range_start_pragma_processor` as safe, we wrapped not only strings but also whole tuples in `Markup()`.
- For tuple elements we didn't differentiate whether they from pygments (already escaped) or from source file (not yet escaped). Only the later must be explicitly escaped.

Relates to #1921.


strictdoc/export/html/generators/source_file_view_generator.py
```diff
@@ -1,5 +1,5 @@
 # mypy: disable-error-code="no-untyped-call,no-untyped-def,operator"
-from typing import List, Tuple
+from typing import List, Tuple, Union
```
stanislaw, line 2: Let's add a new or extend an existing integration test that would contain a good example exercising this change.

haxtibal: I intend to do this. Probably tomorrow. You can wait for it, or go with two separate PRs...

strictdoc/export/html/generators/source_file_view_generator.py
```diff
@@ -154,7 +160,9 @@ def get_pygmented_source_lines(
         pygmented_source_file_content = pygmented_source_file_content[
             slice_start:slice_end
         ]
-        pygmented_source_file_lines = pygmented_source_file_content.split("\n")
+        pygmented_source_file_lines: List[Union[str, RangeMarkerTuple]] = list(
```
haxtibal, line 163: This is for mypy... Honestly I don't understand why we have to make a list()-copy of `pygmented_source_file_content.split("\n")`. Without list(), mypy would claim
```
error: Incompatible types in assignment (expression has type "List[str]", variable has type "List[Union[str, RangeMarkerTuple]]")
```

I know assignment would be a problem without explicit annotation of `pygmented_source_file_lines` because of list invariance. But the annotation is there... why do we still need list()?

strictdoc/export/html/generators/source_file_view_generator.py
stanislaw: I think at this point it would be great to see just one test where all instances of escaping are collected together:

- Escaping a section title
- Escaping a requirement title
- Escaping a requirement field
- Escaping the source line marks
- Diff/Changelog
- (did I miss anything?)

This topic deserves its own folder, let's create one `tests/integration/commands/export/html/escaping` and have that test there.

stanislaw: Please do it in this PR.

haxtibal: Can our filecheck match across newlines? I'm looking for something similar to [[[::space::]]](https://llvm.org/docs/CommandGuide/FileCheck.html#matching-newline-characters) in LLVM filecheck to match this div as a whole
```html
<div
  class="document_title"
  title="&lt;b&gt;&#34;document&#34; &amp; &#39;title&#39;&lt;/b&gt;"
  data-file_name="input.sdoc"
>&lt;b&gt;&#34;document&#34; &amp; &#39;title&#39;&lt;/b&gt;</div>
```

haxtibal: See 83afcc7. How about that approach?

Btw., it already helped in that it revealed surprising behavior. I first used `CHECK-NOT: {{["<>]}}` for the negative check, i.e. including single quote because it's HTML special character. This would fail because single quotes are not escaped in statements. This is because docutils seemingly escapes double quotes, but not single quotes.

stanislaw: Here is an example of the most simplest filecheck matching that I would use: https://github.com/strictdoc-project/strictdoc/compare/stanislaw/fix_sdoc_fileref_itest_example?expand=1. Just explicitly match for what we expect to be escaped. Note that I also slightly extended your test document, adding `document title`, `section title`, etc., to easier visualize which node/field we are dealing with.

FileCheck.py implements only a subset of what LLVM does because I didn't need more over the years and because no one asked. The new filecheck library that we will probably switch to soon will be able to do more.
I prefer to keep the FileCheck use in StrictDoc to most basic features. All the existing test most of the time use this direct matching of the HTML content.

Your custom grep commit is a nice idea but I think we don't need just yet unless I missed something that cannot be achieved using my branch's matching approach?

haxtibal: > Here is an example of the simplest filecheck matching

Thanks, you pointer to `CHECK-NEXT` was important, I didn't know it expects consecutive lines. That's exactly what I had looked for.

> Your custom grep commit is a nice idea but I think we don't need just yet unless I missed something that cannot be achieved using my branch's

It's a trade off. Our tests now expect a bit too much. E.g. markupsafe is free to change their escaping logic to use `&quot;` instead of `&#34;` with their next minor update. Our tests would start to fail, even if escaping is still fine.

And we may add another HTML element somewhere derived from an to-be-escaped value in future. It's not unlikely we would forget to update the test for full coverage.

Advantage of the now exact matches is they'll catch errors in the surrounding HTML context which are strictly speeking not escaping problems, but still are problems. I think this outweighs.

haxtibal:
[This failed MacOS e2e test](https://github.com/strictdoc-project/strictdoc/actions/runs/10130508488/job/28011977903?pr=1929) looks unrelated. I'm not on MacOS, can't reproduce locally. Do you know what's up?

strictdoc/export/html/generators/source_file_view_generator.py
stanislaw: > > Here is an example of the simplest filecheck matching
> 
> Thanks, you pointer to `CHECK-NEXT` was important, I didn't know it expects consecutive lines. That's exactly what I had looked for.

Good. I didn't know what you knew and if you missed anything.

> > Your custom grep commit is a nice idea but I think we don't need just yet unless I missed something that cannot be achieved using my branch's
> 
> It's a trade off. Our tests now expect a bit too much. E.g. markupsafe is free to change their escaping logic to use `&quot;` instead of `&#34;` with their next minor update. Our tests would start to fail, even if escaping is still fine.

I see that your recent commit only uses filecheck for matching, and this is what I would expect here.

I don't see it as a problem if we would have to update in case the escaping logic changes.

> And we may add another HTML element somewhere derived from an to-be-escaped value in future. It's not unlikely we would forget to update the test for full coverage.

This is true. I don't know yet how to deal with this long-term. I think we can only be mindful of the escaping topic from now on. What helps here partially is that we now have the `Markup` type hints, so seeing them very often when working with the StrictDoc code will support the awareness.

> Advantage of the now exact matches is they'll catch errors in the surrounding HTML context which are strictly speeking not escaping problems, but still are problems. I think this outweighs.

Yes, this is how I see it as well. In the past, this kind of redundancy in matching the exact lines helped me several times to identify breaking changes that I was going to introduce.

stanislaw:
> [This failed MacOS e2e test](https://github.com/strictdoc-project/strictdoc/actions/runs/10130508488/job/28011977903?pr=1929) looks unrelated. I'm not on MacOS, can't reproduce locally. Do you know what's up?

It is one of the tests that are known to be flaky. They fail _sometimes_. Most of the time they are timeout errors of some kind but not always. I track the known ones here, and this one didn't fail before: https://github.com/strictdoc-project/strictdoc/issues/1912.

I just restarted the macOS tests and they have all passed.

strictdoc/export/html/generators/source_file_view_generator.py
stanislaw: I added this to the Code climate ticket. I will review this at some point and see if I can resolve it in a different way. For now, low priority. 

https://github.com/strictdoc-project/strictdoc/issues/1912

stanislaw (approved):
Looks good. Thanks for adding the integration tests. I checked both tests' outputs locally, and everything seems to be correct.

I edited your issue to track the remaining items where the other areas have to be double-checked with respect to HTML escaping:

https://github.com/strictdoc-project/strictdoc/issues/1920

PR #1932, opened 2024-08-04
HTML2PDF: System chromedriver for PDF export

haxtibal:
PDF export requires chrome/chromedriver. Currently StrictDoc always uses `webdriver_manager` to download a suitable chromedriver and install in a strictdoc cache subdirectory.

There may be reasons to prefer a system installation over an adhoc download (e.g. security policy). Notably Debian provides packages that work out-of-the-box for StrictDoc. GitHub Ubuntu CI images have the upstream version pre installed.

This adds an CLI option `--chromedriver` to select an explicit chromedriver. If not given, strictdoc uses `webdriver_manager` as usual.

To use the Debian package, one would call
```
apt install chromium chromium-driver
strictdoc export --fromats=html2pdf --chromedriver=/usr/bin/chromedriver .
```

To use chromedriver from GitHub Ubuntu image, one would call
```
 strictdoc export --fromats=html2pdf --chromedriver=$CHROMEWEBDRIVER .
```

### DIFF
--- a/strictdoc/export/html2pdf/html2pdf.py
+++ b/strictdoc/export/html2pdf/html2pdf.py
@@ -164,18 +164,20 @@ def get_pdf_from_html(driver, url) -> bytes:
     return data
 
 
-def create_webdriver():
+def create_webdriver(chromedriver: Optional[str]):
     print("HTML2PDF: creating Chrome Driver service.", flush=True)  # noqa: T201
+    if chromedriver is None:
+        cache_manager = HTML2PDF_CacheManager(
+            file_manager=FileManager(os_system_manager=OperationSystemManager())
+        )
 
-    cache_manager = HTML2PDF_CacheManager(
-        file_manager=FileManager(os_system_manager=OperationSystemManager())
-    )
-
-    http_client = HTML2PDF_HTTPClient()
-    download_manager = WDMDownloadManager(http_client)
-    path_to_chrome = ChromeDriverManager(
-        download_manager=download_manager, cache_manager=cache_manager
-    ).install()
+        http_client = HTML2PDF_HTTPClient()
+        download_manager = WDMDownloadManager(http_client)
+        path_to_chrome = ChromeDriverManager(
+            download_manager=download_manager, cache_manager=cache_manager
+        ).install()
+    else:
+        path_to_chrome = chromedriver
     print(f"HTML2PDF: Chrome Driver available at path: {path_to_chrome}")  # noqa: T201
 
     service = Service(path_to_chrome)
@@ -211,6 +213,11 @@ def main():
     os.environ["WDM_LOCAL"] = "1"
 
     parser = argparse.ArgumentParser(description="HTML2PDF printer script.")
+    parser.add_argument(
+        "--chromedriver",
+        type=str,
+        help="Optional chromedriver path. Downloaded if not given.",
+    )
     parser.add_argument("paths", help="Paths to input HTML file.")
     args = parser.parse_args()
 
@@ -218,7 +225,7 @@ def main():
 
     separate_path_pairs = paths.split(";")
 
-    driver = create_webdriver()
+    driver = create_webdriver(args.chromedriver)
 
     @atexit.register
     def exit_handler():
--- a/docs/strictdoc_01_user_guide.sdoc
+++ b/docs/strictdoc_01_user_guide.sdoc
@@ -2642,7 +2642,10 @@ There are three methods of PDF printing available:
 
 3. Also in the web interface, by navigating to a 'PDF' view of a document and using the browser's built-in Print function.
 
-The first two methods require the Chrome browser to be installed on the user's computer.
+The first two methods require the Chrome browser and chromedriver to be installed on the user's computer.
+StrictDoc downloads chromedriver on demand by default, or uses a pre installed executable if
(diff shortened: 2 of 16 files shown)


haxtibal:
@stanislaw Sorry if these CI debug trials cause E-Mail spam at your side. I try to figure out why $CHROMEWEBDRIVER (defined by GitHub CI image) is not passed into lit in the tox py37 env, while it works in py312 env.

haxtibal:
e2e test `edit_grammar_add_field_move_up_and_down_save` [fails on macOS](https://github.com/strictdoc-project/strictdoc/actions/runs/10244497289/job/28337588222?pr=1932), but works on Linux and Windows. It should be unrelated to this PR, but can't prove it.

stanislaw:
> e2e test `edit_grammar_add_field_move_up_and_down_save` [fails on macOS](https://github.com/strictdoc-project/strictdoc/actions/runs/10244497289/job/28337588222?pr=1932), but works on Linux and Windows. It should be unrelated to this PR, but can't prove it.

Yeah, that seems to be a rather rare flaky test. I recorded it in the code climate ticket. For now, just restart.

> @stanislaw Sorry if these CI debug trials cause E-Mail spam at your side. I try to figure out why $CHROMEWEBDRIVER (defined by GitHub CI image) is not passed into lit in the tox py37 env, while it works in py312 env.

No worries at all! I am happy to see all the updates, including CI. I am glad that you figured it out for LIT.

Overall, your approach looks good but I will see if I have any specific comments.

strictdoc/export/html2pdf/html2pdf.py
```diff
@@ -211,14 +213,19 @@ def main():
     os.environ["WDM_LOCAL"] = "1"
 
     parser = argparse.ArgumentParser(description="HTML2PDF printer script.")
+    parser.add_argument(
```
stanislaw, line 216: Let's document this shortly in the User Guide. Also mention that the Chrome Driver / Browser are used as a printing machine.

haxtibal: How should we support this for `strictdoc server`? The web UI can initiate the pdf export as well, so it would need either the same CLI argument, or it would have to be parsed from strictdoc.toml.

stanislaw: Ideally, it should support both CLI and TOML. But as with everything else, this can be postponed unless you need this right away. Do whatever you have resources for.

stanislaw:
@haxtibal the tests like this one `edit_grammar_move_field_down` are known issues but they used to fail very rarely. I did one round of checking why the could fail and didn't find an obvious reason back then.

I often get an impression that SeleniumBase/Selenium receive updates to their logic and this makes some of these flaky tests to re-surface and start failing more often. Right now it looks like these errors have nothing to do with your branch. I recorded all of these in the code climate and should investigate them, otherwise we will get blocked all the time in the future.

stanislaw:
One reason that used to plague a lot of the end2end tests used to be a generic HTTP 500 Web Driver Exception. A solution to this error in that case was to implement a retry mechanism, catching this 500 error. After that, you can very often see that this machine actually works all the time, catching these 500 errors.

I think this may be the only strategy that actually works: catch and retry, two times.

haxtibal:
> I think this may be the only strategy that actually works: catch and retry, two times.

SeleniumBase has retry logic built in and we see it in action. The [failing exception](https://github.com/strictdoc-project/strictdoc/actions/runs/10263620240/job/28395859472?pr=1932#step:6:464) happens in [this code path](https://github.com/seleniumbase/SeleniumBase/blob/master/seleniumbase/fixtures/base_case.py#L613), which is already the retry of a retry, IIUC.

Any ideas how to reproduce this for debugging, or get good debug information? Ideally we could attach into a running gh runner session, but not sure if that's possible at all. Or maybe making a browser screenshot right before the failing test statement, or dumping the DOM state...

stanislaw:
> > I think this may be the only strategy that actually works: catch and retry, two times.
> 
> SeleniumBase has retry logic built in and we see it in action. The [failing exception](https://github.com/strictdoc-project/strictdoc/actions/runs/10263620240/job/28395859472?pr=1932#step:6:464) happens in [this code path](https://github.com/seleniumbase/SeleniumBase/blob/master/seleniumbase/fixtures/base_case.py#L613), which is already the retry of a retry, IIUC.
> 
> Any ideas how to reproduce this for debugging, or get good debug information? Ideally we could attach into a running gh runner session, but not sure if that's possible at all. Or maybe making a browser screenshot right before the failing test statement, or dumping the DOM state...

I cannot check this right now. What I normally do is to explore the context around the failing line very carefully and try to add intermediate checks or add the retry logic. No silver bullets in my pocket.

Looking at the number of the tests randomly failing, I assume SeleniumBase/Selenium received an update that changed logic/timing and this revealed the existing weaknesses in the tests that are currently failing.

haxtibal:
> I cannot check this right now. What I normally do is to explore the context around the failing line very carefully and try to add intermediate checks or add the retry logic. No silver bullets in my pocket.

I'll revert the last debug commit to set the PR in a reviewable/functional state.

And then will do experiments regarding the failing test in a separate branch or personal repository. [Debug your GitHub Actions by using ssh](https://github.com/marketplace/actions/debugging-with-ssh) sounds interesting.

Issue #1913, opened 2024-07-07
Export: move the --view option from the passthrough to the export command


stanislaw:

stanislaw:
Closed by https://github.com/strictdoc-project/strictdoc/pull/1937.

PR #1937, opened 2024-08-17
Move passthrough commands to export --formats sdoc

haxtibal:
We have two commands for similar tasks:

- `strictdoc export` does sdoc to `<other>` conversion and filtering.
- `strictdoc passthrough` does sdoc to sdoc identity conversion and filtering.

Both commands can be unified by introducing a new export output format `sdoc`. In particular:

```
strictdoc passthrough --view PRINT_VIEW  # old
strictdoc export --view PRINT_VIEW --formats sdoc  # new

strictdoc passtrough --filter-requirements FILTER_REQUIREMENTS  # old
strictdoc export --filter-requirements FILTER_REQUIREMENTS --formats sdoc  # new

strictdoc passthrough --filter-sections FILTER_SECTIONS  # old
strictdoc export --filter-sections FILTER_SECTIONS --formats sdoc  # new
```

Note: Output from `strictdoc passthrough` goes now to `<output-dir>/sdoc`.

Relates to #1913.


haxtibal:
@stanislaw Some questions

- Should all `passthrough` options be supported by `export --formats sdoc`, or only some of them?
- Should we keep the old `passthrough` command and options for backwards compatibility? Mark them as deprecated?
- Should we keep `class PassthroughAction`, or merge it into `class ExportAction`?

stanislaw:
> @stanislaw Some questions
> 
>     * Should all `passthrough` options be supported by `export --formats sdoc`, or only some of them?
> 
>     * Should we keep the old `passthrough` command and options for backwards compatibility? Mark them as deprecated?
> 
>     * Should we keep `class PassthroughAction`, or merge it into `class ExportAction`?

Hi @haxtibal . This is going in the right direction.

Historically, the passthrough command was one of the first commands ever implemented, and its only purpose was to ensure no regressions for .sdoc -> Python -> .sdoc rountrips. Over time it was extended with things like filters or free-text-to-text.

Now it seems reasonable to merge everything from passthrough to export with the following in mind:

- The export to sdoc without any special option shall behave like the current passthrough, giving us the confidence in the correct .sdoc -> Python -> .sdoc rountrips.
- The passthrough CLI command interface should be left for some time for backwards compatibility (I could think of some users still using it). Adding deprecation message to the CLI command already now makes sense. Eventually we will remove the command.
-  The PassthroughAction can be retired completely in favor of the ExportAction.

stanislaw:
Great stuff! Thanks.

PR #1954, opened 2024-10-18
Fix generated search URLs for project statistics

haxtibal:
Search URLs generated for the project statistic page were caught by Jinja2 autoescaping, which makes them invalid.

A good place to mark them as safe is the view object: It's the immediate layer below Jinja templates and specifically made to provide strings included by a Jinja2 template.

Relates to #1920.

### DIFF
--- a/strictdoc/export/html/generators/view_objects/project_statistics_view_object.py
+++ b/strictdoc/export/html/generators/view_objects/project_statistics_view_object.py
@@ -2,6 +2,8 @@
 from dataclasses import dataclass
 from datetime import datetime
 
+from markupsafe import Markup
+
 from strictdoc import __version__
 from strictdoc.core.document_tree_iterator import DocumentTreeIterator
 from strictdoc.core.project_config import ProjectConfig
@@ -34,19 +36,19 @@ class ProjectStatisticsViewObject:
         self.is_running_on_server: bool = project_config.is_running_on_server
         self.strictdoc_version = __version__
 
-    def render_screen(self, jinja_environment: JinjaEnvironment):
+    def render_screen(self, jinja_environment: JinjaEnvironment) -> Markup:
         return jinja_environment.render_template_as_markup(
             "screens/project_statistics/index.jinja", view_object=self
         )
 
-    def render_static_url(self, url: str):
-        return self.link_renderer.render_static_url(url)
+    def render_static_url(self, url: str) -> Markup:
+        return Markup(self.link_renderer.render_static_url(url))
 
-    def render_url(self, url: str):
-        return self.link_renderer.render_url(url)
+    def render_url(self, url: str) -> Markup:
+        return Markup(self.link_renderer.render_url(url))
 
-    def render_static_url_with_prefix(self, url: str):
-        return self.link_renderer.render_static_url_with_prefix(url)
+    def render_static_url_with_prefix(self, url: str) -> Markup:
+        return Markup(self.link_renderer.render_static_url_with_prefix(url))
 
     def is_empty_tree(self) -> bool:
         return self.document_tree_iterator.is_empty_tree()
--- a/strictdoc/export/html/generators/view_objects/document_screen_view_object.py
+++ b/strictdoc/export/html/generators/view_objects/document_screen_view_object.py
@@ -246,7 +246,7 @@ class DocumentScreenViewObject:
             else "/".join((root_prefix, document_link))
         )
 
-    def render_static_url(self, url: str) -> str:
+    def render_static_url(self, url: str) -> Markup:
         return Markup(self.link_renderer.render_static_url(url))
 
     def render_local_anchor(self, node) -> str:
--- a/strictdoc/export/html/templates/components/table_key_value/index.jinja
+++ b/strictdoc/export/html/templates/components/table_key_value/index.jinja
@@ -7,7 +7,11 @@
     <div class="sdoc-table_key_value-section">{{ obj["Section"] }}</div>
   {% else %}
     {% if view_object.project_config.is_activated_search() and obj["Link"] is defined %}
-      <a class="sdoc-table_key_value-key" href="{{ obj["Link"] }}">{{ obj["Key"] }}</a>
+      <a
+        class="sdoc-table_key_value-key"
+        data-testid="search-{{ obj["Key"].lower().replace(" ", "-") }}"
+        href="{{ obj["Link"] }}">{{ obj["Key"] }}
+      </a>
     {% else %}
       <div class="sdoc-table_key_value-key">{{ obj["Key"] }}</div>
     {% endif %}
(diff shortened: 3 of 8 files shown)


haxtibal:
Unrelated to this MR: The new e2e test show two results that I wouldn't have expected

> Root-level requirements not connected to by any requirement 0

This always remains 0, no matter where I place unconnected requirements. What is it meant to count?

> Non-root-level requirements not connected to any parent requirement 4

Should IMO be 5. It seems that `node["STATUS"] != "Backlog"` evaluates to `False` if `STATUS` is not defined. I would expect logic like `undefined STATUS != "Backlog" == True`. Just like NaN ≠ x is always True.

haxtibal:
`render_url()` and friends should internally do URL escaping with something like `urllib.parse.urlencode`, then mark as safe. This is currently not considered, but we could add it to the TODO list in #1920.

strictdoc/export/html/generators/view_objects/project_statistics_view_object.py
```diff
@@ -34,19 +36,19 @@ def __init__(
...
 
-    def render_screen(self, jinja_environment: JinjaEnvironment):
+    def render_screen(self, jinja_environment: JinjaEnvironment) -> Markup:
         return jinja_environment.render_template_as_markup(
             "screens/project_statistics/index.jinja", view_object=self
         )
 
-    def render_static_url(self, url: str):
-        return self.link_renderer.render_static_url(url)
+    def render_static_url(self, url: str) -> Markup:
```
stanislaw, line 44: Would it make sense to move the `Markup(...)` down to the link_renderer? Because that's where that content originates?
This comment applies to all occurrences of this change.

stanislaw (approved):

stanislaw:
> Unrelated to this MR: The new e2e test show two results that I wouldn't have expected
> 
> > Root-level requirements not connected to by any requirement 0
> 
> This always remains 0, no matter where I place unconnected requirements. What is it meant to count?

A root-level requirement becomes 'connected' if there is another requirement that has a RELATION to it.

> > Non-root-level requirements not connected to any parent requirement 4
> 
> Should IMO be 5. It seems that `node["STATUS"] != "Backlog"` evaluates to `False` if `STATUS` is not defined. I would expect logic like `undefined STATUS != "Backlog" == True`. Just like NaN ≠ x is always True.

This logic was written with the assumption that the Backlog requirements do not contribute to these project statistics. If a requirement has no status, it can then be found with another metric and its status should be set to a meaningful vallue. I think it is pretty consistent or?

stanislaw:
> `render_url()` and friends should internally do URL escaping with something like `urllib.parse.urlencode`, then mark as safe. This is currently not considered, but we could add it to the TODO list in #1920.

Added!

strictdoc/export/html/generators/view_objects/project_statistics_view_object.py
stanislaw: I left this comment but let's discuss it independently or this MR which I will merge now as-is.

haxtibal:
> A root-level requirement becomes 'connected' if there is another requirement that has a RELATION to it.

Yes but the query says  "_not_ connected to". Nothing in the test input.sdoc is connected (there are no relations), so this should be 5, not 0. The question comes down to: How would `node.is_root` as standalone query become `True`? For me it evaluates to `False` even for requirements directly written under `[DOCUMENT]`.

> This logic was written with the assumption that the Backlog requirements do not contribute to these project statistics. If a requirement has no status, it can then be found with another metric and its status should be set to a meaningful vallue. I think it is pretty consistent or?

Not sure. All other queries in project statistics are written so that backlog requirements _do_ contribute. That's a bit inconsistent.

Anyway, I made a wrong assumption in my previous comment:

>  It seems that node["STATUS"] != "Backlog" evaluates to False 

It actually evaluates to True and that's good.

strictdoc/export/html/generators/view_objects/project_statistics_view_object.py
haxtibal: > move the Markup(...) down to the link_renderer

The "problem" is, that `Markup` is "infectious". Any `str` that's somehow mixed (+, replace, ...) with `Markup` will itself become `Markup` and gets escaped on the way. This is not always what we want. Lower level functions may produce output that's not only meant to end up in HTML, but also into other formats. HTML escaping would be inadequate there, but it's hard to control if Markup objects are unexpectedly passed around.

My conclusion until now is that returning `Markup` is good, but only if we can guarantee that a function is only ever used in HTML context.

stanislaw:
> > A root-level requirement becomes 'connected' if there is another requirement that has a RELATION to it.
> 
> Yes but the query says "_not_ connected to". Nothing in the test input.sdoc is connected (there are no relations), so this should be 5, not 0. The question comes down to: How would `node.is_root` as standalone query become `True`? For me it evaluates to `False` even for requirements directly written under `[DOCUMENT]`.

Now I understand what was missing. The document needs to be marked as ROOT: True, see this for example:

```
[DOCUMENT]
MID: b7bcca934cd148c3855beb4c85723bc5
TITLE: Requirements Tool Specification (L1)
REQ_PREFIX: SDOC-SSS-
ROOT: True
OPTIONS:
  ENABLE_MID: True
```

I found no better way of assigning a root because StrictDoc cannot know what is a root of a graph.

> > This logic was written with the assumption that the Backlog requirements do not contribute to these project statistics. If a requirement has no status, it can then be found with another metric and its status should be set to a meaningful vallue. I think it is pretty consistent or?
> 
> Not sure. All other queries in project statistics are written so that backlog requirements _do_ contribute. That's a bit inconsistent.

Yes, I can see a bit of consistency. The idea was to exclude Backlog reqs clearly from the `not connected to`-stats because otherwise you could end up with a non-zero number for these stats even if the non-backlog requirements are all traced.

Which / are there any other metrics you would want to see with `!= Backlog` too?

> Anyway, I made a wrong assumption in my previous comment:
> 
> > It seems that node["STATUS"] != "Backlog" evaluates to False
> 
> It actually evaluates to True and that's good.

Good.

haxtibal:
> Now I understand what was causing the confusion. The document needs to be marked as ROOT: True

Now I understand too :) Thanks for explaining. Just learnt one can mark even multiple documents with `ROOT: True`. This makes sense in our use case. We have multiple documents that can be considered a possible source (= root?) of requirements

- external requirements from IEC, connected via a compliance matrix (as not all external requirements apply, it could make sense to consider the compliance matrix document as root)
- requirements to the operating system
- requirements to the application

> Which / are there any other metrics you would want to see with `!= Backlog` too?

Up to now we don't use the `Backlog` state. But we have things like a "Proposed" state for REQUIREMENT and for custom grammar and have not defined metrics yet. Guess it's hard to find a "fits everywhere" statistics page. Maybe it would make sense to be able to define a custom statistics page (with custom queries)?

PR #1993, opened 2024-11-10
Mark diff colors safe and escape raw diff input

haxtibal:
For HTML escaping of the diff view we have to consider two things.

1. Diff input comes from two git checkouts of the project at specific revisions. The revisions sdocs are considered untrusted user input, could contain special characters and must be escaped.
2. After analyzing with difflib we add a bit HTML to colorize the output. This specific HTML fragments are trusted and safe.

Relates to #1920.

### DIFF
--- a/strictdoc/export/html/generators/view_objects/diff_screen_view_object.py
+++ b/strictdoc/export/html/generators/view_objects/diff_screen_view_object.py
@@ -2,6 +2,8 @@
 from dataclasses import dataclass
 from datetime import datetime
 
+from markupsafe import Markup
+
 from strictdoc import __version__
 from strictdoc.core.project_config import ProjectConfig
 from strictdoc.export.html.html_templates import JinjaEnvironment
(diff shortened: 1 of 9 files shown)


strictdoc/git/project_diff_analyzer.py
```diff
@@ -481,14 +481,14 @@ def _iterate_one_index(
                             if node.title != other_section_or_none.title:
                                 title_modified = True
                                 lhs_colored_title_diff = (
-                                    get_colored_diff_string(
+                                    get_colored_html_diff_string(
```
stanislaw, line 484: Let's give the using class the `Markup` type, i.e.,

```
        self.lhs_colored_title_diff: Optional[Markup] = lhs_colored_title_diff
        self.rhs_colored_title_diff: Optional[Markup] = rhs_colored_title_diff
```

strictdoc/export/html/generators/view_objects/diff_screen_view_object.py
```diff
@@ -39,27 +41,27 @@ def __init__(
...
             "screens/git/index.jinja", view_object=self
         )
 
-    def render_url(self, url: str):
-        return self.link_renderer.render_url(url)
+    def render_url(self, url: str) -> Markup:
+        return Markup(self.link_renderer.render_url(url))
 
-    def render_node_link(self, incoming_link, document, document_type):
+    def render_node_link(self, incoming_link, document, document_type) -> str:
```
stanislaw, line 52: Could you explain why some render methods are Markup and some are str?

Is there a way we could unify this?

haxtibal: Honestly I simply copied the signatures from document_screen_view_object.py because I remember back then I had a reason for `-> str`. But now I don't remember, so it's good you ask, we should reconsider. You could help me by telling when this method is invoked at all. I searched but didn't find a caller.

stanislaw: 1) I was wondering if all these render* methods should return `Markup`? If there is no specific reason to ever have `str`, we would have it consistent.

2) `render_node_link` could be a copy-and-paste from the other view object methods. Just remove it because I also don't see it being used.

stanislaw: Actually, let me merge this one, and let's handle the str/Markup/unused stuff separately.

PR #2011, opened 2024-11-21
Feature: Export Doxygen TAGFILE with a map from requirements to StrictDoc documentation

stanislaw:


johanenglund:
Oops, I just noticed this after writing #2013.

PR #2016, opened 2024-11-22
doxygen: Change compound type and file suffix

johanenglund:
Change compound type from page to file to prevent that all tags end up in the doxygen TOC. 
Change export file suffix from .xml to .tag.


stanislaw:
Thanks!

PR #2034, opened 2024-11-30
test/integration: add custom script for exporting questionnaires

BenGardiner:
This is a custom export script and sdoc documents as input that we will use in https://github.com/nmfta-repo/nmfta-telematics_security_requirements (we proved-out use of sdoc in https://github.com/nmfta-repo/vcr-experiment/ before that).

Please let me know what changes I can apply here to make this useful to you -- if at all. Rejecting the PR is also completely fine. 😊


stanislaw (commented):
Hi @BenGardiner, thanks a lot for contributing this is such a clean way. I am almost ready to merge it but just two comments:

1\) Would you be open to also contributing a small tutorial, explaining your use case, referencing your script and making a couple of examples? I have been thinking for a long time about this new genre of StrictDoc documentation, and that is Tutorials 💥 💥 💥 . Your tutorial could be an excellent start into this, explaining what a user of StrictDoc can do with it using Python. If you are up for it, you could create a small tutorial in a dedicated SDoc file in the docs/tutorials/ folder.

1b) In my world, I am not used to the term "questionnaire" when it comes to answering questions about requirements. What I and other people (speaking from I have seen in aerospace) are used to is called "requirements compliance". Is my understanding right that you will send this questionnaire to users, and they give their answers? If so, there are some examples of different flavors of this compliance matrix topic:

- https://www.google.com/url?sa=t&source=web&rct=j&opi=89978449&url=https://www.safran-group.com/download/media/441641&ved=2ahUKEwip3PqPi4eKAxW6RPEDHYSKOHMQFnoECBoQAQ&usg=AOvVaw0iIWVmZj_CpRON4QOqleTZ  (this one features Compliant (C), Non-compliant (NC), Partiall Compliant (PC), and Not Applicable (N/A)

- https://lws-set.gsfc.nasa.gov/documents/SEARS_App_A_Experime_77CE4A.pdf (this one also features the verification types column where each requirement has to be verified according to a certain method, such as review, analysis, inspection or test and others.

(**Your questionnaire template is perfectly fine, I just wanted to mention the templates that I am used to seeing and working with.**)


2\) Is there a chance you could reduce your SDoc files to just a small number of requirements necessary to demonstrate the operation of the script? Maintaining this code long-term does not need to have all the SDoc content in my opinion. You could create just a few requirements, each highlighting something that you script does in various cases.

I am very happy that you are working on this.

BenGardiner:
Awesome, glad it could be useful.

> 1) Would you be open to also contributing a small tutorial, explaining your use case, referencing your script and making a couple of examples? I have been thinking for a long time about this new genre of StrictDoc documentation, and that is Tutorials 💥 💥 💥 . Your tutorial could be an excellent start into this, explaining what a user of StrictDoc can do with it using Python. If you are up for it, you could create a small tutorial in a dedicated SDoc file in the docs/tutorials/ folder.

Sure I can work on that.

> Is my understanding right that you will send this questionnaire to users, and they give their answers?

The fleets can give them to their current and potential suppliers of telematics along with a request to complete. Then the suppliers respond, yes. 

I agree, compliance is a good word for this too. 

> If so, there are some examples of different flavors of this compliance matrix topic:

Thanks! These are something to aspire-to ; very thorough.

> 2) Is there a chance you could reduce your SDoc files to just a small number of requirements necessary to demonstrate the operation of the script? Maintaining this code long-term does not need to have all the SDoc content in my opinion. You could create just a few requirements, each highlighting something that you script does in various cases.

yeah that is a totally reasonable request. I started with this goal as well for this PR but fell-back to the whole contents because I notice that we needed all four fragments because the script uses them by name, then I would need to cook up 2 names then we needed at least 3 fragement hosted requirement with each of the parent requirement criticalities and then also requirements that are included in some fragments and not others and then also the custom grammar... it just seemed easier to keep the sdocs as-is rather than hunt around for all these special.

stanislaw:
> yeah that is a totally reasonable request. I started with this goal as well for this PR but fell-back to the whole contents because I notice that we needed all four fragments because the script uses them by name, then I would need to cook up 2 names then we needed at least 3 fragement hosted requirement with each of the parent requirement criticalities and then also requirements that are included in some fragments and not others and then also the custom grammar... it just seemed easier to keep the sdocs as-is rather than hunt around for all these special.

Yeah, I will merge it as-is for now because your example is self-contained. Later I could maybe revisit it a reduce it some but it is a minor issue for now.

PR #2052, opened 2025-01-03
Added 'child-relation' to diff view for enhanced change impact visualization

mplum:
Added 'child-relation' to the requirement diff view.

image


stanislaw:
Thanks a lot!

PR #2069, opened 2025-01-27
Use rST admonitions where possible for "Note", "Observation", "Example"

nogjam:
Noticed as I was using/editing the docs that we could clean some things up by using rST directives.
Here's one example of the change implemented:

**Before:**

image

**After:**

image


stanislaw:
Thanks! I appreciate your attention to this detail.

Issue #2099, opened 2025-03-04
Bug: Exception if same file location is referenced several times


haxtibal:
## Describe the bug

Referencing the same source code location with multiple markers causes an exception. See reproducer.

### Reproducer 1: Forward and backward reference to same function

__document.sdoc__
```
[DOCUMENT]
TITLE: Reproducer

[REQUIREMENT]
UID: REQ-1
TITLE: Requirement 1
STATEMENT: REQ-1 references forward to file.c:foo by file relation.
RELATIONS:
- TYPE: File
  VALUE: file.c
  FUNCTION: foo

[REQUIREMENT]
UID: REQ-2
TITLE: Requirement 2
STATEMENT: file.c:foo references "backward" to REQ-2 by function annotation in C comment.
```

__file.c__
```c
// @relation(REQ-2, scope=function)
void foo() { }
```

```
$ strictdoc export .
...
AssertionError: (SourceMarkerTuple(before_line=Markup('// @relation('), after_line=Markup(', scope=function)'), marker=FunctionRangeMarker(parent = NoneType(...), reqs_objs = [1 elements], reqs = [1 elements], scope = RangeMarkerType(...), ng_source_line_begin = 1, ng_source_column_begin = NoneType(...), ng_range_line_begin = 1, ng_range_line_end = 2, ng_marker_line = 1, ng_marker_column = 14, ng_is_nodoc = False, _description = "function foo")), <class 'str'>)
```

Note: The same would work for python functions. Difference is, C function markers are generated so that comment lines above a function are part of the function line range.

### Reproducer 2: Two forward references to same location

__document.sdoc__
```
[DOCUMENT]
TITLE: Reproducer

[REQUIREMENT]
UID: REQ-1
TITLE: Requirement 1
STATEMENT: REQ-1 references forward to file.c:foo by file relation.
RELATIONS:
- TYPE: File
  VALUE: file.c
  FUNCTION: foo

[REQUIREMENT]
UID: REQ-2
TITLE: Requirement 2
STATEMENT: REQ-1 references forward to file.c:foo and file.c:foo bar by line range.
RELATIONS:
- TYPE: File
  VALUE: file.c
  LINE_RANGE: 1, 3
```

__file.c__
```c
void foo() { }

void bar() { }
```

```
$ strictdoc export .
...
[...]

PR #2104, opened 2025-03-07
Support roles for file relations and markers

haxtibal:
Some incomplete snippets of what will be possible.

Forward file relations with roles...
```
[REQUIREMENT]
UID: REQ-001
...
RELATIONS:
- TYPE: File
  ROLE: Implementation
  VALUE: foo.py
  FUNCTION: foo
```

and "backward" file relations with roles
```py
@pytest.mark.req("@relation(REQ-002, scope=line, role=Test)")
def test_some_function_of_req002():
    some_test_assertion()
```

See integration tests for more examples.

This is WIP. I created the MR anyway to enable discussions. Known TODOs:
- [x] Forward class relation not working. See also #2113.
- [x] ~Better strategy for multiple markers per SourceMarkerTuple needed.~ Will be done with #2099.
- [ ] I don't like the special handling of simple forward file references in map_reqs_uids_to_paths_with_role; we should better create dynamic markers for it, as already done for forward function/class/range relations.
- [x] Section "Resolve definitions to declarations (only applicable for C and C++)" in validate_and_resolve creates a marker without role. Have to understand the use case first to adapt this part.
- [x] 2 failing integrations tests with Python 3.8, something about calling `__init__()` with wrong args
- [ ] Role is only rendered in document view, not yet in "special views".
- [X] Let tests test, not just render. Note: Only input.html, but not source code html is checked, because #2099 will likely change the output (see "req badges").
- [ ] Only trivial test cases are provided; special cases and error handling missing.
- [ ] User Guide.

To execute test cases:
```
invoke test-integration --focus 01_relations_with_role
invoke test-integration --focus 02_forward_relations_with_role
```

### DIFF
--- a/strictdoc/core/file_traceability_index.py
+++ b/strictdoc/core/file_traceability_index.py
@@ -1,6 +1,7 @@
 # mypy: disable-error-code="arg-type,attr-defined,no-any-return,no-untyped-def"
 from typing import Any, Dict, Iterator, List, Optional, Set, Tuple, Union
 
+from strictdoc.backend.sdoc.error_handling import StrictDocSemanticError
 from strictdoc.backend.sdoc.models.node import SDocNode
 from strictdoc.backend.sdoc.models.reference import FileReference
 from strictdoc.backend.sdoc.models.type_system import FileEntry
@@ -32,22 +33,27 @@ class FileTraceabilityIndex:
         # "file.py" -> List[SDocNode]
         self.map_paths_to_reqs: Dict[str, OrderedSet[SDocNode]] = {}
 
-        # "REQ-001" -> {"file.py", ...}
-        self.map_reqs_uids_to_paths: Dict[str, OrderedSet[str]] = {}
+        # "REQ-001" -> {("file.py", "Implementation"), ...}
+        self.map_reqs_uids_to_paths_with_role: Dict[
+            str, OrderedSet[Tuple[str, Optional[str]]]
+        ] = {}
 
         # "file.py" -> SourceFileTraceabilityInfo
         self.map_paths_to_source_file_traceability_info: Dict[
             str, SourceFileTraceabilityInfo
         ] = {}
 
+        # "REQ-1" -> [ ("file.py", (10, 12), "Impl" ), ... ]
         self.map_reqs_uids_to_line_range_file_refs: Dict[
-            str, List[Tuple[str, Tuple[int, int]]]
+            str, List[Tuple[str, Tuple[int, int], Optional[str]]]
         ] = {}
+
+        # "file.py" -> { { "foo" -> [("REQ-1", "Impl"), ("REQ-2", "Test")] }, ... }
         self.map_file_function_names_to_reqs_uids: Dict[
-            str, Dict[str, List[str]]
+            str, Dict[str, List[Tuple[str, Optional[str]]]]
         ] = {}
         self.map_file_class_names_to_reqs_uids: Dict[
-            str, Dict[str, List[str]]
+            str, Dict[str, List[Tuple[str, Optional[str]]]]
         ] = {}
 
         # This is only public non-static functions from languages like C.
@@ -77,22 +83,25 @@ class FileTraceabilityIndex:
 
     def get_requirement_file_links(
         self, requirement: SDocNode
-    ) -> List[Tuple[str, Optional[List[RangeMarker]]]]:
-        if requirement.reserved_uid not in self.map_reqs_uids_to_paths:
+    ) -> List[Tuple[str, Optional[str], Optional[List[RangeMarker]]]]:
+        if (
+            requirement.reserved_uid
+            not in self.map_reqs_uids_to_paths_with_role
+        ):
             return []
 
         matching_links_with_opt_ranges: List[
-            Tuple[str, Optional[List[RangeMarker]]]
+            Tuple[str, Optional[str], Optional[List[RangeMarker]]]
         ] = []
-        requirement_source_paths: OrderedSet[str] = self.map_reqs_uids_to_paths[
-            requirement.reserved_uid
-        ]
+        requirement_source_paths: OrderedSet[Tuple[str, Optional[str]]] = (
+            self.map_reqs_uids_to_paths_with_role[requirement.reserved_uid]
+        )
 
         # Now that one requirement can have multiple File-relations to the same file.
         # This can be multiple FUNCTION: or RANGE: forward-relations.
         # To avoid duplication of results, visit each unique file link path only once.
         visited_file_links: Set[str] = set()
-        for requirement_source_path_ in requirement_source_paths:
+        for requirement_source_path_, forward_role in requirement_source_paths:
             if requirement_source_path_ in visited_file_links:
                 continue
             visited_file_links.add(requirement_source_path_)
@@ -111,11 +120,11 @@ class FileTraceabilityIndex:
             )
             if not markers:
                 matching_links_with_opt_ranges.append(
-                    (requirement_source_path_, None)
+                    (requirement_source_path_, forward_role, None)
                 )
                 continue
             matching_links_with_opt_ranges.append(
-                (requirement_source_path_, markers)
+                (requirement_source_path_, forward_role, markers)
             )
 
         return matching_links_with_opt_ranges
@@ -189,8 +198,6 @@ class FileTraceabilityIndex:
         # STEP: Collect minimal information that will help to resolve the
         #       forward-declared paths/function names at the step 2.
         #
-        marker_type: RangeMarkerType
-
         for trace_info_ in self.trace_infos:
             source_file: SourceFile = assert_cast(
                 trace_info_.source_file, SourceFile
@@ -290,9 +297,9 @@ class FileTraceabilityIndex:
                     file_posix_path, OrderedSet()
                 ).add(forward_requirement_)
 
-                self.map_reqs_uids_to_paths.setdefault(
+                self.map_reqs_uids_to_paths_with_role.setdefault(
                     forward_requirement_.reserved_uid, OrderedSet()
-                ).add(file_posix_path)
+                ).add((file_posix_path, relation_.role))
 
                 if file_reference.g_file_entry.function is not None:
                     one_file_function_name_to_reqs_uids = (
@@ -302,7 +309,9 @@ class FileTraceabilityIndex:
                     )
                     one_file_function_name_to_reqs_uids.setdefault(
                         file_reference.g_file_entry.function, []
-                    ).append(forward_requirement_.reserved_uid)
+                    ).append(
+                        (forward_requirement_.reserved_uid, relation_.role)
+                    )
                 elif file_reference.g_file_entry.clazz is not None:
                     one_file_class_name_to_reqs_uids = (
                         self.map_file_class_names_to_reqs_uids.setdefault(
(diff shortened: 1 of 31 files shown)


haxtibal:
Hi @stanislaw, I don't understand the rationale behind the `if not function_.is_public():` case in FileTraceabilityIndex.validate_and_resolve. Could you explain?

I can see how it works for the public (i.e., extern) case:
```c
// @relation(REQ-001, scope=function, role=Implementation)
extern int extfoo();

// Understood: strictdoc will automatically add a marker derived from the above declaration marker to the definition here 
int extfoo() {
    return 0;
}
```

But the same does intentionally not work for the private (i.e., static) case:
```c
// @relation(REQ-001, scope=function, role=Implementation)
static int foo();

// Why?: strictdoc will *not* automatically add a marker derived from the above declaration marker to the definition here 
static int foo() {
    return 0;
}
```

Why are extern and static handled differently?

stanislaw:
> Hi @stanislaw, I don't understand the rationale behind the `if not function_.is_public():` case in FileTraceabilityIndex.validate_and_resolve. Could you explain?
> 
> Why are extern and static handled differently?

This is most probably just me who didn't think through this case and didn't handle it 🤔 

Feel free to handle this like it should be but I would say, open a separate issue for that and don't get side-tracked by it while working on the file relations roles. Otherwise the PR will grow too much in terms of the semantics involved.

Another note: 

When handling the JUnit XML and Google Test macros, I had to move some code around. Specifically, to simplify the pretty complex code, I moved all the linking inside file_traceability_index to just one function and that is `validate_and_resolve`. I am not changing what the code does, just moving all previously scattered linking chunks to one place. The method is now even bigger than it was before but the advantage is that it is possible to follow through all the validation/resolution steps in one place.

This will definitely cause merge conflicts but I can support you if you get confused about the change. But I think it should be a little more consolidated now and could be further refactored later on.

stanislaw:
The FileTraceabilityIndex is a pseudo-graph database class that models something like this. Originally it started with just a couple of dictionaries and then grew larger. I am looking into smarter ways of managing these links.

One recent thought is to split the FileTraceabilityIndex into two: FileTraceabilityIndex (just the data containers) and FileTraceabilityIndexBuilder where the algorithmic things like `validate_and_resolve` will be contained.

Furthermore, I considered making FileTraceabilityIndex to have a graph-like API and even considered NetworkX. However, I don't see a good way forward. This is something that we could discuss during the office hours, maybe you would have ideas about managing all this nicely.

The key challenge is that it is not possible to populate graph in parts but there is a transaction where some things are added first, then some other things based on them, then some validations and final linking and fixups. I guess now that you encountered this complexity yourself you should know exactly which kind of code organization challenge I mean.

Note: The diagram is not entirely accurate but it gives an idea of the graph we are dealing with.

Screenshot 2025-03-19 at 23 25 02

stanislaw:
I would also add that **because** this code has become a mess, it takes much longer to implement new stuff and makes it easy to make mistakes.

> Why are extern and static handled differently?

In fact, I had this somewhere in the back of my mind and wanted to double-check but then I most probably missed because I was focusing on dealing with the other code handling details. If the code was cleaner, we would have a better time developing it further.

haxtibal:
> In fact, I had this somewhere in the back of my mind and wanted to double-check but then I most probably missed because I was focusing on dealing with the other code handling details.

No problem. I've already fixed the file role related bit about it. Besides that, yes, the if/else looks like you had something in mind. So I better don't interfere with it.

strictdoc/core/file_traceability_index.py
```diff
@@ -496,10 +511,12 @@ def create_requirement(
                     )
                     if overwrite:
                         assert (
-                            requirement.reserved_uid
-                            not in function_name_to_reqs_uids
-                        )
-                    function_name_to_reqs_uids.append(requirement.reserved_uid)
+                            requirement.reserved_uid,
+                            file_reference.role,
```
haxtibal, line 515: @stanislaw

> This will definitely cause merge conflicts but I can support you

Resolving conflicts was relatively easy. You still may want to double check I didn't mess up your intended semantics.

One particular place I noted is here. You were checking for duplicated UIDs in function_name_to_reqs_uids and class_name_to_reqs_uids. After my change, an entry is only considered duplicated if both UID and role match. So now it would be allowed to have to entries like `[("REQ1", None), ("REQ1", "TestCase")]` there.

strictdoc/core/file_traceability_index.py
```diff
@@ -433,16 +424,39 @@ def marker_comparator(marker):
...
+        def path_with_role_comparator(path_with_role):
+            return path_with_role[0]
 
         # Sort by keys alphabetically.
-        for paths_ in self.map_reqs_uids_to_paths.values():
-            paths_.sort()
+        for paths_with_role in self.map_reqs_uids_to_paths_with_role.values():
+            paths_with_role.sort(key=path_with_role_comparator)
 
     def create_requirement(
```
stanislaw, line 461: Note that the create_requirement was renamed to `create_requirement_with_forward_source_links` and it now looks like this:

```
    def create_requirement_with_forward_source_links(
        self, requirement: SDocNode
    ) -> None:
        self.requirements_with_forward_links.add(requirement)
```

all the logic was moved to be a separate step in `validate_in_resolve`.

strictdoc/core/file_traceability_index.py
stanislaw: > One particular place I noted is here. You were checking for duplicated UIDs in function_name_to_reqs_uids and class_name_to_reqs_uids. After my change, an entry is only considered duplicated if both UID and role match. So now it would be allowed to have to entries like [("REQ1", None), ("REQ1", "TestCase")] there.

This makes sense.

strictdoc/helpers/ordered_set.py
```diff
@@ -48,5 +48,13 @@ def __str__(self) -> str:
     def __repr__(self) -> str:
         return f"<OrderedSet {self}>"
 
-    def sort(self) -> None:
-        self._d = dict(sorted(self._d.items()))
+    def sort(self, key=None) -> None:
```
stanislaw, line 51: I didn't implement this in my branch, so my comparator function when sorting OrderedSet was giving me tuples, not objects. You will have to fix this in one place here:

```
        # Sort by node UID alphabetically.
        for path_requirements_ in self.map_paths_to_reqs.values():

            def node_comparator(node) -> int:
                return assert_cast(node[0].reserved_uid, str)

            path_requirements_.sort(key=node_comparator)
```

strictdoc/core/file_traceability_index.py
stanislaw: All the lines should be changed in `validate_in_resolve`, so there should be not much room for surprises.

haxtibal:
@stanislaw There is a [integration test failing](https://github.com/strictdoc-project/strictdoc/actions/runs/13966806004/job/39098968884?pr=2104#step:7:418) for (only) python3.8

```
tox -e py38-check -- lit ...
...
Traceback (most recent call last):
  File "/Users/runner/work/strictdoc/strictdoc/strictdoc/backend/sdoc_source_code/reader_c.py", line 359, in read_from_file
    sdoc = self.read(sdoc_content, file_path=file_path)
  File "/Users/runner/work/strictdoc/strictdoc/strictdoc/backend/sdoc_source_code/reader_c.py", line 57, in read
    py_language = Language(  # type: ignore[call-arg, unused-ignore]
TypeError: __init__() missing 1 required positional argument: 'name'
```

caused by this code

```py
# Works since Python 3.9 but we also lint this with mypy from Python 3.8.
language_arg = tree_sitter_cpp.language()
py_language = Language(  # type: ignore[call-arg, unused-ignore]
    language_arg
)
```

Not sure, but I think we had this already, it was about a tree sitter problem with python 3.8? Do you remember? If so, how are integration tests intended to pass?

stanislaw:
> @stanislaw There is a [integration test failing](https://github.com/strictdoc-project/strictdoc/actions/runs/13966806004/job/39098968884?pr=2104#step:7:418) for (only) python3.8
> 
> ```
> tox -e py38-check -- lit ...
> ...
> Traceback (most recent call last):
>   File "/Users/runner/work/strictdoc/strictdoc/strictdoc/backend/sdoc_source_code/reader_c.py", line 359, in read_from_file
>     sdoc = self.read(sdoc_content, file_path=file_path)
>   File "/Users/runner/work/strictdoc/strictdoc/strictdoc/backend/sdoc_source_code/reader_c.py", line 57, in read
>     py_language = Language(  # type: ignore[call-arg, unused-ignore]
> TypeError: __init__() missing 1 required positional argument: 'name'
> ```
> 
> caused by this code
> 
> ```python
> # Works since Python 3.9 but we also lint this with mypy from Python 3.8.
> language_arg = tree_sitter_cpp.language()
> py_language = Language(  # type: ignore[call-arg, unused-ignore]
>     language_arg
> )
> ```
> 
> Not sure, but I think we had this already, it was about a tree sitter problem with python 3.8? Do you remember? If so, how are integration tests intended to pass?

Look for this line in the tests:

```
REQUIRES: PYTHON_39_OR_HIGHER
```

stanislaw:
FYI: I was thinking of dropping the Python 3.8 support soon'ish, also due to this tree-sitter minimal requirement. I think 3.8 is already quite old.

stanislaw:
One test has failed due to https://github.com/strictdoc-project/strictdoc/issues/2121. I don't know yet what happens on Windows but there is something strange going on. Needs to be reproduced and investigated on Windows. For now, I just restarted that job.

haxtibal:
> Furthermore, I considered making FileTraceabilityIndex to have a graph-like API

Yes, such an API could be helpful. As a contributor I should understand your mental model. E.g. what you consider to be a node, an edge, a node attribute. It's currently a bit hidden since we operate directly on a bunch of dicts of dicts of lists and so on. If we had instead something like `db.add_node()`, `db.add_edge()`, ... the model would become more obvious.

> even considered NetworkX

Looked at NetworkX for a project some years ago, then opted for igraph (because of the C backend). I think API wise they're up to par. We used them as in memory DAG structure and wrapped that into some python classes with domain specific API. We didn't call it "graph DB". Maybe because I simply don't know what turns a graph into a graph DB :-) Can you explain? There must be a difference, since there are also dedicated graph DBs (like [cogdb](https://cogdb.io/)).

> However, I don't see a good way forward. This is something that we could discuss during the office hours, maybe you would have ideas about managing all this nicely.

For me it would be 1. learning (about the current model and graph DBs in general), 2. some isolated prototypes. Then reconsider. You have more experience in that area and maybe know shortcuts.

> The key challenge is that it is not possible to populate graph in parts but there is a transaction where some things are added first, then some other things based on them, then some validations and final linking and fixups.

I don't see a problem with that multiple passes (or stages or steps, however you call it). We could maybe make them more explicit. And we could make the data that is passed between stages but not required afterwards short lived (not 100% sure, but trace_infos and requirements_with_forward_links seem to fall into that category). Having them as FTI attributes gives the impression they're needed to operate the DB, but actually they're only needed to initially build the DB.

> The diagram is not entirely accurate but it gives an idea of the graph we are dealing with.

IIUC you consider a UID string "REQ-001" to be not only a node-attribute (of SDocNode), but at same time also a node on its own - is that correct?

haxtibal:
Ok, all tests pass and functionality is about what I had initially intended. I'd say the PR has reached a minimal mergeable state (except of upcoming review comments that need to be fixed, of course).

Now the question is
- try to merge soon, to avoid further conflicts?
- or try to enhance more (see open TODOs)?

strictdoc/export/html/generators/source_file_view_generator.py
```diff
@@ -231,9 +230,26 @@ def get_pygmented_source_lines(
         for marker in coverage_info.markers:
             marker_line = marker.ng_source_line_begin
             assert isinstance(marker_line, int)
-            pygmented_source_file_line = assert_cast(
-                pygmented_source_file_lines[marker_line - 1], str
-            )
+            # TODO: Don't merge, will get a dedicated fix with #2099
```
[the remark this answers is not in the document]
stanislaw: If I understand correctly, your PR is mergable without this part of the changeset. Please remove this change and we can already merge this PR to integrate the increment.

@mettta and I will most likely work on https://github.com/strictdoc-project/strictdoc/issues/2099 (the UI and the related fix) tomorrow, so it should happen independently.

stanislaw (approved):
Excellent!

PR #2123, opened 2025-03-21
WIP: Represent forward file reference as marker

haxtibal:
strictdoc handled the simple forward file relation listed below special:

```
[REQUIREMENT]
UID: REQ-001
TITLE: Requirement Title 1
STATEMENT: >>>
Requirement with forward relations.
<<<
RELATIONS:
- TYPE: File
  VALUE: file.py
```

All other file references (to functions, classes, lines, ranges, file backwards) are represented as markers internally. Only the plain forward file reference was not a marker, but an entry to a separate dict.

Let's resolve the simple forward file relation to a range marker covering the whole file (1st line to last line) and get rid of the special handling. This is consistent with what we get when we use a @relation(..., scope=file) backward marker.

TODO:
- [ ] tests of course explode, since the generated source file html is now different (covering 1st to last line); will adjust once the approach is aggreed
- [ ] to be discussed: instead representing as range marker, we could also introduce a new FileMarker and ForwardFileMarker

### DIFF
--- a/strictdoc/backend/sdoc_source_code/models/range_marker.py
+++ b/strictdoc/backend/sdoc_source_code/models/range_marker.py
@@ -48,7 +48,7 @@ class RangeMarker:
         self.ng_range_line_begin: Optional[int] = None
         self.ng_range_line_end: Optional[int] = None
 
-        self.ng_is_nodoc = "nosdoc" in self.reqs
+        self.ng_is_nodoc: bool = "nosdoc" in self.reqs
         self.ng_new_relation_keyword = scope is not None and len(scope) > 0
 
     def is_begin(self) -> bool:
@@ -74,8 +74,8 @@ class LineMarker:
     ) -> None:
         assert isinstance(reqs_objs, list)
         self.parent = parent
-        self.reqs_objs = reqs_objs
-        self.reqs = list(map(lambda req: req.uid, reqs_objs))
+        self.reqs_objs: List[Req] = reqs_objs
+        self.reqs: List[str] = list(map(lambda req: req.uid, reqs_objs))
         self.role: Optional[str] = (
             role if role is not None and len(role) > 0 else None
         )
@@ -87,7 +87,7 @@ class LineMarker:
         self.ng_range_line_begin: Optional[int] = None
         self.ng_range_line_end: Optional[int] = None
 
-        self.ng_is_nodoc = "nosdoc" in self.reqs
+        self.ng_is_nodoc: bool = "nosdoc" in self.reqs
 
     def is_begin(self) -> bool:
         return True
@@ -113,7 +113,7 @@ class ForwardRangeMarker:
         assert len(reqs_objs) > 0
         self.start_or_end: bool = start_or_end
 
-        self.reqs_objs = reqs_objs
+        self.reqs_objs: List[str] = reqs_objs
         self.role: Optional[str] = (
             role if role is not None and len(role) > 0 else None
         )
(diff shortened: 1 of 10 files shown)


stanislaw:
> All other file references (to functions, classes, lines, ranges, file backwards) are represented as markers internally. Only the plain forward file reference was not a marker, but an entry to a separate dict.

A clear result of a continuous evolution. The plain forward file reference was the first one ever implemented and it didn't need the markers. Now that we have all sorts of entities, you are making a very good point of unifying the approach 👍 

> Let's resolve the simple forward file relation to a range marker covering the whole file (1st line to last line) and get rid of the special handling. This is consistent with what we get when we use a @relation(..., scope=file) backward marker.

👍 

> tests of course explode, since the generated source file html is now different (covering 1st to last line); will adjust once the approach is aggreed

The approach is agreed. Please proceed.

> to be discussed: instead representing as range marker, we could also introduce a new FileMarker and ForwardFileMarker

I would say I am leaning towards a "yes" on this. Just one comment: at some point, I was wondering if it was possible to refactor all markers to just one class Marker and avoid maintaining polymorphism but this, if it ever happens, would be a separate work item, so let's implement a proper FileMarker/ForwardFileMarker for now.

P.S.: The source view screen UI work is WIP. Will probably share a good screenshot or a branch this week still.

haxtibal:
> > to be discussed: instead representing as range marker, we could also introduce a new FileMarker and ForwardFileMarker
> 
> I would say I am leaning towards a "yes" on this. Just one comment: at some point, I was wondering if it was possible to refactor all markers to just one class Marker and avoid maintaining polymorphism but this, if it ever happens, would be a separate work item, so let's implement a proper FileMarker/ForwardFileMarker for now.

Doing that right now.

I wonder whether we should display such FileMarker/ForwardFileMarker like a range, because we could think of it as a range that covers the whole file, from 1st to last line?

Or should we present it special - as it was before, where the source code view said "See requirements that refer to this source file without specifying range", and no lines were highlighted?

stanislaw:
> Doing that right now.
> 
> I wonder whether we should display such FileMarker/ForwardFileMarker like a range, because we could think of it as a range that covers the whole file, from 1st to last line?
> Or should we present it special - as it was before, where the source code view said "See requirements that refer to this source file without specifying range", and no lines were highlighted?

We are still working on the Source View screen redesign (see the screenshot, but expect quite a few UI changes). The direction we are going is as follows:

- The left panel's previous approach "with/without range requirements" is gone. Now the top level items will be ranges/markers. Each marker will have children items requirements, each having references to files/ranges, including the files which are not the current file.
- The source view: above each source line there will be three ways of displaying relevant markers associated with that line: 1\) just a number of markers associated (minimal view), 2\) some information about the markers (medium view), and 3\) the full requirement templates (detailed view, similar to what can be seen on the screenshot).

We are looking into redesigning the source view screen in a way that we can optionally access requirements with their full content as well as only provide a minimalistic view that does not distract from reading the source code only if a user wants it.

We are struggling a little as to what should be the best UI solution but we are on it.

What this means for you:

I would create a FileMarker/ForwardFileMarker just like the other markers. It must have a range (begin and end), so that it can be treated like any other marker. This way these whole file-markers will always have ranges (1, length of the file) and this is how the new UI will highlight them as such.

Bildschirmfoto 2025-03-23 um 20 17 12

stanislaw:
This work has been integrated as part of #2126. Thanks for contributing it @haxtibal !

Issue #2107, opened 2025-03-09
Feature: JUnit XML -> SDoc verification report generator


stanislaw:
## Description

This feature is a building block for these two general feature discussions:

- https://github.com/strictdoc-project/strictdoc/discussions/1640
- https://github.com/strictdoc-project/strictdoc/issues/2009

## Problem

We need some way to integrate the verification reports back to the StrictDoc documentation.

JUnit XML is one of the most common formats that the test tools produce. We could test the proof of concept with JUnit XML and then do the same for https://ctrf.io/ and other formats.

## Solution

Write a Python script that reads both SDoc and source files and the JUnit XML. In addition to resolving the normal relations between SDoc and source files, the script shall also parse the JUnit XML information and generate a SDoc verification report that contains verification items linking to the test source files as well as the original requirements.

## Additional Information

If a general JUnit XML script is possible (it should be), more scripts can be written and integrated to the StrictDoc repository in the same way.

Open question that can be handled when the script is in place:

- Should the script be directly integrated to the `strictdoc` tool or kept as standalone script? My answer: yes, should be integrated.

stanislaw:
cc @johanenglund @nogjam @haxtibal @mplum @i-interprete-on-the-fly @joergmis

stanislaw:
An extra challenge here seems to be that JUnit XML format is used slightly differently depending on a test tool and level type:

Here's a fragment from what is produced by PyTest (StrictDoc's unit tests). Here the paths are dot-separated folder/file path components. The names are source code function names.

```xml
<?xml version="1.0" encoding="utf-8"?>
<testsuites>
    <testsuite name="pytest" errors="0" failures="0" skipped="0" tests="239" time="4.150" timestamp="2025-03-09T12:22:00.027102+01:00" hostname="home">
        <testcase classname="tests.unit.strictdoc.backend.sdoc.models.test_requirement" name="test_01_empty_requirement_as_starting_point" time="0.004" />
        <testcase classname="tests.unit.strictdoc.backend.sdoc.models.test_requirement" name="test_02_mutating_with_a_field" time="0.006" />
        <!-- ... -->
    </testsuite>
</testsuites>
```

Here's a fragment from what is produced by LLVM LIT (StrictDoc's integration tests). Here the classname is a combination of an ID + an incomplete path to a test file. The name is just a test file name.

```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuites time="372.58">
[...]

PR #2205, opened 2025-05-01
commands/manage_autouid: ensure that non-SDoc files are not modified

stanislaw:
https://github.com/strictdoc-project/strictdoc/issues/2107#issuecomment-2844991711

### DIFF
--- a/strictdoc/commands/manage_autouid_command.py
+++ b/strictdoc/commands/manage_autouid_command.py
@@ -70,6 +70,11 @@ class ManageAutoUIDCommand:
                 next_number += 1
 
         for document in traceability_index.document_tree.document_list:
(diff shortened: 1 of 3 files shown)


stanislaw:
cc @johanenglund

johanenglund:
This fix works for me. I did however run into another issue on the main branch when testing this, it fails to copy my image assets when doing a html export. Any change here that I've missed?

stanislaw:
> This fix works for me. I did however run into another issue on the main branch when testing this, it fails to copy my image assets when doing a html export. Any change here that I've missed?

We haven't changed anything with regard to the assets. Could you open another issue and provide details/stacktrace?

PR #2265, opened 2025-05-20
Code climate: add custom fixit linting check: StrictDoc_CheckDocstringsRule

stanislaw:


stanislaw:
cc @thseiler @haxtibal 

It looks like fixit is pretty usable for this kind of stuff. I don't have too much time to make this perfect right now, but it already does pretty well on the docstrings. Another check could be added normal `#`-comments, reusing some of the helpers developed with this PR.

stanislaw:
A small disadvantage of this is that it is pretty slow compared to how fast Ruff format and linter is.

PR #2267, opened 2025-05-21
Code climate: add custom fixit linting check: StrictDoc_CheckCommentsRule

stanislaw:


stanislaw:
cc @haxtibal @thseiler 

It is a super quick and dirty implementation that together this the previous PR does:

```
invoke lint-fixit
...
🛠️  223 files checked, 56 files with errors, 124 auto-fixes available 🛠️ 
```

I will merge this now and handle the edge cases over time. Thanks @haxtibal for the hint about fixit2!

thseiler:
Very nice!

haxtibal:
Nice to see this in action :+1:

Ruff has some related rules, e.g. [Checks for docstrings that do not start with a capital letter](https://docs.astral.sh/ruff/rules/first-word-uncapitalized/) and [Checks for docstrings in which the first line does not end in a period.](https://docs.astral.sh/ruff/rules/missing-trailing-period/#missing-trailing-period-d400). Just mentioning to double check we don't write custom rules for what's already in ruff.

stanislaw:
> Nice to see this in action 👍
> 
> Ruff has some related rules, e.g. [Checks for docstrings that do not start with a capital letter](https://docs.astral.sh/ruff/rules/first-word-uncapitalized/) and [Checks for docstrings in which the first line does not end in a period](https://docs.astral.sh/ruff/rules/missing-trailing-period/#missing-trailing-period-d400). Just mentioning to double check we don't write custom rules for what's already in ruff.

Yes, thanks for mentioning those. I did review those and some other recently but found that they were not doing exactly what this is doing now. There may be some overlap but it is not a complete one.

stanislaw:
One other rule I was thinking of would be to ban the use of @property in the codebase completely because it can lead to some confusing debates in one's mind as to whether something should be written as property or a normal get_() function. Right now, StrictDoc has several inconsistent uses of @property and I was thinking of getting rid of them because anyway most of the time the normal functions do the job and there are way more of them.

Meaning, banning @properties and only using:

- Instance variables created in __init__
- Normal def functions for everything else.

PR #2268, opened 2025-05-21
code-climate: mypy: address type-arg

stanislaw:


stanislaw:
@thseiler I am integrating your trivial changes here and need to consider the other less trivial changes separately.

PR #2269, opened 2025-05-21
code-climate: mypy: address type-arg: SourceFileTraceabilityInfo/Parts

stanislaw:


stanislaw:
@thseiler I had isolated your Part-related change in this PR, and then I started having doubts about that approach.

First I thought that we could simply create a TypeAlias to a Union, similar to how I do it for `SDocElementIF`.

But then I also realized that we don't even use `g_parts`. The only reason why it is used in the constructor is that TextX can proxy through it through different markers and parse them through the `general_language_marker_processors`. To summarize: I think, we could put Any and do nothing in this constructor and related types because `g_parts` is never used from this Python class.

PR #2270, opened 2025-05-22
code-climate: mypy: address type-arg: JSON generator

stanislaw:


strictdoc/export/json/json_generator.py
```diff
@@ -218,7 +232,7 @@ def get_level_string_(node_) -> str:
             return subnode_dict
 
         elif isinstance(node, SDocDocument):
-            node_dict: Dict[str, List[Dict]] = {JSONKey.NODES: []}
```
stanislaw, line 221: @thseiler A few lines above I am getting:
Bildschirmfoto 2025-05-22 um 09 39 02

I believe that you would hit this error right after you re-enable further mypy checks in this file.

stanislaw: In general, you generalized the types to the JSON structure which is very interesting.

I am wondering if some of the declarations, such as `section_dict` could be even more type-specific and instead of a general `Dict[str, JSONType]:` in the above example you would directly tell it that we have a dictionary of dictionaries of JSONType or similar. To constraint it even more where we clearly know that the generator has specific final types.

stanislaw: I will explore doing it this evening.

stanislaw: I could not make it work with all mypy checks disabled and I regressed into a basic `Dict[str, Any]` approach. I don't think it is worth fighting for given this code is pretty simple and encapsulated.

Issue #2231, opened 2025-05-08
Feature: Support for parsing C/C++ test descriptions


Master5:
## Description

Add support for parsing C/C++ test descriptions independently of any specific test framework. The parser should extract structured test documentation embedded in comments above test functions, using recognizable markers such as @test and Markdown headers.

## Problem

Currently, there is no built-in support for parsing test descriptions directly from C/C++ source files without relying on a specific test framework (e.g., CTest, GoogleTest, etc.). This makes it difficult to extract metadata or structured documentation from tests in a consistent way across different codebases.

Additionally, without structured parsing, it's not possible to automatically generate software test documentation that includes traceability links (e.g., to requirements or test cases). This limits the ability to ensure compliance with standards or maintain a clear test-to-requirement mapping.

## Solution

Implement a general-purpose C/C++ parser that can identify specially formatted comment blocks placed above test functions. These blocks should include markers like @test and structured headers e.g. **Expected Results**:.

To support downstream tooling, the format of these headers should align with the grammar supported by StrictDoc, allowing seamless generation of documentation and traceability reports. This enables the automatic creation of structured test documentation compatible with existing documentation pipelines or compliance tools.

```
/** 
 * @test Test Initialization
 *
 * @relation(TC-00001, scope=function)
 * 
 * **Intention**: blabla  
 *
 * **Input**:  
 * blabla  
 *
 * **Expected Results**:  
 * blabla
 */
TEST_CASE("Test Initialization", testInitialization)
{
    ...
}
```

## Additional Information

- The parser should be independent of any specific test framework.
- Could be useful for generating test documentation

nicpappler:
Generally I like the syntax. What I'm just wondering, do we also want a unique ID for a test, so that we can uniquely identify it? I just assume there will not be a separate test case specification in a requirement style describing the test, right? So the test here is both test case specification and test case implementation, right?

johanenglund:
This feature world be a very welcomed addition.

Anything I don't entirely understand the linkage to the tested requirement if TC-00001 in the example is a test case?

stanislaw:
Thanks to everyone for your comments so far. Let me also share my thoughts and considerations.

[...]

PR #2272, opened 2025-05-25
backend/sdoc_source_code: parse SDoc nodes from source files

stanislaw:
This addresses https://github.com/strictdoc-project/strictdoc/issues/2231 for C/C++.

1) Lark parser is introduced for parsing code comments. Besides `@relation` markers, now the arbitrary nodes can be parsed as well, for example:

```
/**
 * Some text.
 *
 * @relation(REQ-1, scope=function)
 *
 * INTENTION: Intention...
 *
 * INPUT: This
 *        is
 *        a statement
 *        \n\n
 *        And this is the same statement's another paragraph.
 *
 * EXPECTED_RESULTS: This is a comment.
 */
void test_case_1(void) {
    print("hello world\n");
}
```

2) Using `CAPITALIZED_NAME: <...>` was the most straightforward both in terms of parsing and interfacing with the existing SDoc convention where the field names are capitalized. I tested `\n\n` to work successfully for creating paragraphs in both Doxygen and SDoc HTML.

3) Note that the strictdoc.toml has this important config option now:

```
source_nodes = [
  { "tests/" = { uid = "TEST_DOC", node_type = "TEST_SPEC" } }
]
```

...which gives a hint to the source node parser where to inject the auto-generated nodes and which type they should be (in this case description.sdoc has TEST_DOC and defines a grammar with SECTION and TEST_SPEC).

4) I was not sure how to auto-generate sections, so I went with sorting them according to the source code function paths.

The feature is super-raw but I prefer to merge it and be testing the integration early.

Related to: #2231

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,6 +49,8 @@ classifiers = [
 # @sdoc[SDOC-SRS-89]
 dependencies = [
     "textx >= 4.0.0, == 4.*",
+    "lark >= 1.2.2",
+
     "jinja2 >= 2.11.2",
     # Reading project config from strictdoc.toml file.
     "toml",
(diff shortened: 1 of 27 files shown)


stanislaw:
cc @thseiler @haxtibal @Master5

PR #2298, opened 2025-06-01
[DISCARDED] Code climate: export_action: simplify building of traceability index

stanislaw:


stanislaw:
@thseiler I tried to make a small change but failed at it because it requires more changes that I don't want to go into now.

Your PR left it in a decent shape anyway, so I will close this for now until I have a clear idea on how to move forward.

PR #2305, opened 2025-06-05
[[NODE]] migration: enhance the Diff screen to highlight diffs for all node types

stanislaw:


stanislaw:
OMG/ReqIF site is down, so there is one dead link. Merging anyway since everything else has passed.

PR #2333, opened 2025-06-15
parallelizer: fix the shutdown edge case on Windows 

### LINKED ISSUE #2083: Parallellization edge case on Windows, tests/integration/features/html/parallelization/06_parallelization_sdoc_parsing_problem_in_child_process

stanislaw:
Finally closes https://github.com/strictdoc-project/strictdoc/issues/2083.

### DIFF
--- a/strictdoc/helpers/parallelizer.py
+++ b/strictdoc/helpers/parallelizer.py
@@ -18,6 +18,7 @@ def map_does_not_work(self, contents, processing_func):
         return executor.map(processing_func, contents)
 """
 
+import atexit
 import multiprocessing
 import sys
 import traceback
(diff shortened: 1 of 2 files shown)


stanislaw:
@thseiler I was too fast with my wishful thinking and closed the issue yesterday but the issue appeared again. I became upset and finally set up the Windows machine. After some tinkering and talking with Copilot about Windows, I found that these two lines fix the issue.

Along the way I became more motivated to finally solve this code climate issue where the SDReader catches some exceptions and does `sys.exit(1)` by itself whereas the better practice would be to only exit from one place and that is from the `main()` function. I will remove this chunk of code pretty soon and all exceptions will be handled at the top level:

```
        except StrictDocException as exception:
            print(f"error: {exception.args[0]}")  # noqa: T201
            sys.exit(1)
        except NotImplementedError:
            traceback.print_exc()
            sys.exit(1)
        except StrictDocSemanticError as exc:
            print(exc.to_print_message())  # noqa: T201
            sys.exit(1)
        except Exception as exc:  # pylint: disable=broad-except
            print(  # noqa: T201
                f"error: could not parse file: "
                f"{file_path}.\n{exc.__class__.__name__}: {exc}"
            )
            # TODO: when --debug is provided.
            traceback.print_exc()
            sys.exit(1)
```

PR #2335, opened 2025-06-16
tests/integration: redirect all build artifacts to build/tests_integration

stanislaw:
This helps to avoid the PyCharm IDE continuously re-indexing the integration tests folder as the itests are run. Having a dedicated build folder for the test artifacts in a cleaner way of building the temporary test artifacts.

### DIFF
--- a/strictdoc/helpers/parallelizer.py
+++ b/strictdoc/helpers/parallelizer.py
@@ -103,35 +103,46 @@ class MultiprocessingParallelizer(Parallelizer):
         # @sdoc[/SDOC_IMPL_2]
 
     def shutdown(self) -> None:
+        bad_child_exit_code: bool = False
+
         # @sdoc[SDOC_IMPL_2]
         # macOS edge case: If the __init__ fails to initialize itself, we may
         # end up having no self.processes attribute at all.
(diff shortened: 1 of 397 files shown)


strictdoc/helpers/parallelizer.py
```diff
@@ -103,35 +103,46 @@ def __init__(self) -> None:
...
 
-        for _ in self.processes:
-            self.input_queue.put(None)
-
         if was_fully_initialized:
+            for _ in self.processes:
+                self.input_queue.put(None)
+
+            for process_ in self.processes:
+                process_.join(timeout=5)
```
stanislaw, line 119: @thseiler I believe this join with timeout fixes the failing test because the parent process does not get blocked forever and then it has an opportunity to terminate the child explicitly because it misbehaves.

This job shows how this works with a hanging child which cannot be joined:

```
error: Parallelizer: Process PID 6200 failed to join within timeout.
```

https://github.com/strictdoc-project/strictdoc/actions/runs/15715994038/job/44285745779

I will keep it at this solution for now because the parent waits and then terminates the child. What is still unclear is why the child would get stuck.

I also confirm that I did see this issue on both 3.9 and 3.12 failing CI builds.

PR #2388, opened 2025-07-20
/backend/sdoc: remove the TITLE-STATEMENT-DESCRIPTION-CONTENT validation

stanislaw:
There are cases where we want to allow none of the standard SDoc reserved fields to be present on a node, even though this means that no SDoc naming conventions are followed.

In particular, some ReqIF files contain fields that follow neither ReqIF nor SDoc naming conventions.

Currently, the fields TITLE, STATEMENT, DESCRIPTION, or CONTENT are used to determine whether a field should be treated as single-line or multiline. Under the new behavior, if none of these fields are present, all fields will be treated as multiline by default.

In the future, we may introduce a separate field for the GrammarElement field declarations, for example, to explicitly specify whether a field should be interpreted as Title, Meta, Content, etc.

### DIFF
--- a/strictdoc/backend/sdoc/models/grammar_element.py
+++ b/strictdoc/backend/sdoc/models/grammar_element.py
@@ -10,7 +10,6 @@ from strictdoc.backend.sdoc.models.model import (
     RequirementFieldName,
 )
 from strictdoc.helpers.auto_described import auto_described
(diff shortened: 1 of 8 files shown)


stanislaw:
cc @thseiler @haxtibal

PR #2498, opened 2025-10-04
feat(project_config): add a new option for reading configuration from a Python file instead of TOML

stanislaw:
Configuring StrictDoc with a Python file provides more flexibility when a user has to:

- Create custom generators
- Remove the existing generators
- Customize requirement and document tree validations
- Customize project statistics

When the Python config feature becomes stable, the TOML config will be deprecated and removed from StrictDoc after some time.


stanislaw:
@thseiler @thseiler, as discussed with both of you at different times during the office hours, I am moving to a Python config incrementally. I have not documented the change yet, because want to give it some testing first. You should get an idea of the new approach by looking at how StrictDoc's own `strictdoc_config.py` looks like now.

The TOML file will go away because it is much easier to provide more advanced configurations with Python directly. Also, the Python config is type-aware, and the IDE can give more information about the possible options.

Issue #2068, opened 2025-01-26
2025: Code climate


stanislaw:

stanislaw:
@thseiler @haxtibal 

I have finally fixed the most important issues related to collecting code coverage from StrictDoc's child processes and got a good number of 94%, see the comment above ☝ .

Several thoughts/observations:

- The code coverage is collected from all 5 test groups as per the `test_all` Invoke task. Running the tests with code coverage enabled makes them considerably slower. I am still to measure it but I am feeling like this combined coverage calculation should not be part of the normal CI. I still think that setting a custom link check that the combined coverage must not fall below a given limit, e.g., 95%, 96%, etc., over time.
- I will be working through some of the most obvious fixes from now on. The next good goal will be to reach 95%.
- Pushing to higher coverage is mostly related to resolving all sorts of non-covered edge cases which opens an interesting trade-off: **if the higher coverage is set as a goal, then all future contributions must become 100% covered as well which increases the amount of work every contributor like you will have to do.**. Previously, we only required mostly green-case integration/end2end test but now much more effort by any contributor would be required.

Do you have any thoughts regarding this topic?

stanislaw:
```
--- Captured Uvicorn stderr ---

INFO:     Started server process [12295]
INFO:     Waiting for application startup.
INFO:     

+========================================================================+
|                      StrictDoc Web Server v0.29.0                      |
|                                                                        |
| Server URL: http://127.0.0.1:5112/                                      |
|                                                                        |
| Documentation: https://strictdoc.readthedocs.io/                       |
|                                                                        |
| Subscribe to the StrictDoc mailing list for news about features,       |
| breaking changes, and other updates:                                   |
| https://groups.io/g/strictdoc                                          |
|                                                                        |
| Share feedback or report issues:                                       |
| https://github.com/strictdoc-project/strictdoc/issues                  |
+========================================================================+

INFO:     Application startup complete.
INFO:     Uvicorn running on http://127.0.0.1:5112/ (Press CTRL+C to quit)
INFO:     127.0.0.1:50010 - "WebSocket /ws/1789317981666" [accepted]
INFO:     connection open
INFO:     127.0.0.1:50012 - "WebSocket /ws/1789317982351" [accepted]
INFO:     connection open
INFO:     Shutting down
ERROR:    Exception in ASGI application
Traceback (most recent call last):
[...]

Issue #2067, opened 2025-01-26
2025: Documentation train


stanislaw:
Previous train: https://github.com/strictdoc-project/strictdoc/issues/1911
Next train: TBD

---

- [x] https://github.com/strictdoc-project/strictdoc/pull/2356
- [ ] Conventional commits.
- [ ] User guide: MultipleChoice vs Tag. See https://github.com/strictdoc-project/strictdoc/pull/2273#issuecomment-2909144366.
- [ ] Testing: SDoc's own documentation. SDoc's reqs in SDoc.
- [ ] Design issue: single vs multiline fields. Define them in grammar instead of relying on the content field idx? https://github.com/strictdoc-project/strictdoc/discussions/2221
- [ ] ReqIF / New default profile.
- [ ] Two workflows: CLI and UI.
- [ ] Composable documents: assets path resolution.
- [ ] FAQ: Related tools: Capella, Gaphor.
- [ ] ReqIF: Export/Import Wizard.
- [ ] ReqIF: describe scripting examples.
- [ ] FAQ: Does StrictDoc have to only have the text markup?
- [ ] Design document: Dealing with empty fields.
- [ ] https://github.com/strictdoc-project/strictdoc/issues/1353, RST markup basic examples. Related: https://asciiflow.com/#/
- [ ] https://www.hackster.io/jens6151/bicycle-computer-on-spresense-b0e332
- [ ] 304 handling (assets and docs)
- [ ] Why feature toggles? (edge cases)
- [ ] `The SDoc in-memory model is converted to an Excel XLSX file using the
XlsWriter library` (end with a dot).

## L1/L2

- [x] Single-line source file links.
- [x] Parsing source nodes. 
- [x] Delete node -> delete_requirement.py
- [x] Traceability matrix ->  traceability_matrix_screen.js
- [x] Tool identification feature L1/L2 ->  about_command.py, version_command.py
- [ ] L1: Distribution requirement: static/PyInstaller/binary.
- [ ] L1: `1.4. Target audience`, create subsections, remove bold titles.
- [ ] Docker L1/L2
- [ ] Level == None L2 requirement.
- [ ] L2: Buttons to copy text to buffer - JS file.
- [ ] Diff feature ->  diff_command.py , many itests
- [ ] Project statistics ->  document_stats.py
- [ ] Search / Query engine -> query_engine/
  - [ ] Static HTML search bar vs Search screen.
- [ ] Doxygen generator ->  doxygen_generator.py
[...]

PR #2527, opened 2025-10-24
Add sanity checks for source_nodes configuration

haxtibal:
Please check if following assumptions are correct.

Using source_nodes depends on having `REQUIREMENT_TO_SOURCE_TRACEABILITY` and `SOURCE_FILE_LANGUAGE_PARSERS` enabled. Further, if custom `include_source_paths` are used, they must include the directories listed in `source_nodes`. Listing directories in in source_nodes is not enough. Warn if one of the above preconditions are not met, and enhance documentation accordingly.

### DIFF
--- a/docs/strictdoc_01_user_guide.sdoc
+++ b/docs/strictdoc_01_user_guide.sdoc
@@ -2987,6 +2987,14 @@ To enable SDoc node parsing, the corresponding source file must be registered in
 
 .. code-block::
 
+    features = [
+      "REQUIREMENT_TO_SOURCE_TRACEABILITY",
+      "SOURCE_FILE_LANGUAGE_PARSERS",
+    ]
+
+    # if custom include_source_paths are set, it must include the corresponding source file
+    include_source_paths = ["tests/**"]
+
     source_nodes = [
         { "tests/" = { uid = "TEST_DOC", node_type = "TEST_SPEC" } }
     ]
--- a/strictdoc/core/project_config.py
+++ b/strictdoc/core/project_config.py
@@ -749,6 +749,14 @@ class ProjectConfigLoader:
 
             if "source_nodes" in project_content:
                 source_nodes_config = project_content["source_nodes"]
+                if len(source_nodes_config) > 0 and not {
+                    ProjectFeature.REQUIREMENT_TO_SOURCE_TRACEABILITY,
+                    ProjectFeature.SOURCE_FILE_LANGUAGE_PARSERS,
+                }.issubset(project_features):
+                    print(  # noqa: T201
+                        "warning: defining source_nodes without enabling REQUIREMENT_TO_SOURCE_TRACEABILITY and "
+                        "SOURCE_FILE_LANGUAGE_PARSERS has no effect"
+                    )
                 assert isinstance(source_nodes_config, list)
                 for item_ in source_nodes_config:
                     source_node_path = next(iter(item_))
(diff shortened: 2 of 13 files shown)


docs/strictdoc_01_user_guide.sdoc
```diff
@@ -2987,6 +2987,14 @@ To enable SDoc node parsing, the corresponding source file must be registered in
 
 .. code-block::
 
+    features = [
+      "REQUIREMENT_TO_SOURCE_TRACEABILITY",
+      "SOURCE_FILE_LANGUAGE_PARSERS",
```
stanislaw, line 2992: This is correct as of today but soon, maybe still in 2025 Q4, I was planning to make SOURCE_FILE_LANGUAGE_PARSERS the default behavior and show a warning that this option is no longer needed.

docs/strictdoc_01_user_guide.sdoc
```diff
@@ -2987,6 +2987,14 @@ To enable SDoc node parsing, the corresponding source file must be registered in
 
 .. code-block::
 
+    features = [
```
stanislaw, line 2990: I have been switching all commits to follow Conventional commits: https://www.conventionalcommits.org/en/v1.0.0/.

So in this case, I would do something like:

```
feat(backend/sdoc_source_code): add sanity checks for source_nodes configuration
```

(I need to update the DevGuide about this, taken a note here: https://github.com/strictdoc-project/strictdoc/issues/2067. In the future, a linter for this could be added).

strictdoc/core/project_config.py
```diff
@@ -748,6 +748,14 @@ def _load_from_dictionary(
             assert section_behavior in ("[SECTION]", "[[SECTION]]")
 
             if "source_nodes" in project_content:
+                if not {
```
stanislaw: I would add two integration tests to not let the code coverage fall down a bit. I have been trying to reach 97% branch coverage recently, and it is quite close (`invoke qualification` generates it btw).

I would add a `_validation` folder to this folder: `tests/integration/features/source_code_traceability/_source_nodes` and there add two tests, following the example of: `tests/integration/features/source_code_traceability/_validation`.

strictdoc/core/project_config.py
```diff
@@ -748,6 +748,14 @@ def _load_from_dictionary(
...
+                    ProjectFeature.REQUIREMENT_TO_SOURCE_TRACEABILITY,
+                    ProjectFeature.SOURCE_FILE_LANGUAGE_PARSERS,
+                }.issubset(project_features):
+                    print(  # noqa: T201
+                        "WARNING: defining source_nodes without enabling REQUIREMENT_TO_SOURCE_TRACEABILITY and "
```
stanislaw: I think we want to always do the lowercase `warning:` moving forward. I am considering this with regard to node validation, and since errors are always displayed as `errors:`, it seems reasonable to be doing `warning:` to match the capitalization.

I did use WARNING: for deprecation messages but now in retrospective I think I was simply too wishful/anxious, wanting the users to migrate ASAP not asap 😄 After a cleanup (separate PR, of course, taking a note in #2068), we could make them to be `warning:` as well.

stanislaw (commented):
This is a nice change that makes implicit things explicit 👍 

I think it is correct as it stands, no edge cases I can think of.

docs/strictdoc_01_user_guide.sdoc
stanislaw: Added a note here to actually make it the default: https://github.com/strictdoc-project/strictdoc/issues/2068.

strictdoc/core/project_config.py
haxtibal: Done

PR #2532, opened 2025-10-27
feat(backend/sdoc_source_code): tighten lark grammar for custom source nodes

haxtibal:
With source node parsing enabled, the comment parser considered **any word** followed by ":" as StrictDoc relevant custom tag. This is overly greedy.

We know right on startup which tag values are allowed and now narrow the search to exactly the allowed values. This prevents parsing errors or unexpected results where comments have "someword:" by chance instead of as intentional SDoc annotation.

@stanislaw: This is just a quick patch to ask whether you would like the idea. If so, I can write tests and think about better design (in particular, passing TraceabilityIndex as a whole to SourceFileTraceabilityCachingReader is a bit too much).

### DIFF
--- a/strictdoc/backend/sdoc_source_code/caching_reader.py
+++ b/strictdoc/backend/sdoc_source_code/caching_reader.py
@@ -4,6 +4,7 @@
 
 from typing import Optional, Union
 
+from strictdoc.backend.sdoc.models.grammar_element import GrammarElement
 from strictdoc.backend.sdoc.pickle_cache import PickleCache
 from strictdoc.backend.sdoc_source_code.models.source_file_info import (
     SourceFileTraceabilityInfo,
@@ -26,7 +27,9 @@ from strictdoc.core.project_config import ProjectConfig
 class SourceFileTraceabilityCachingReader:
     @staticmethod
     def read_from_file(
-        path_to_file: str, project_config: ProjectConfig
+        path_to_file: str,
+        project_config: ProjectConfig,
+        source_node_grammar_element: Optional[GrammarElement],
     ) -> Optional[SourceFileTraceabilityInfo]:
         unpickled_content = PickleCache.read_from_cache(
             path_to_file, project_config, "source_file"
@@ -39,7 +42,7 @@ class SourceFileTraceabilityCachingReader:
             return unpickled_content
 
         reader = SourceFileTraceabilityCachingReader._get_reader(
-            path_to_file, project_config
+            path_to_file, project_config, source_node_grammar_element
         )
         try:
             traceability_info = reader.read_from_file(path_to_file)
@@ -59,7 +62,9 @@ class SourceFileTraceabilityCachingReader:
 
     @staticmethod
     def _get_reader(
-        path_to_file: str, project_config: ProjectConfig
+        path_to_file: str,
+        project_config: ProjectConfig,
+        source_node_grammar_element: Optional[GrammarElement],
     ) -> Union[
         SourceFileTraceabilityReader,
         SourceFileTraceabilityReader_Python,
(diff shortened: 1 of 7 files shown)


strictdoc/backend/sdoc_source_code/caching_reader.py
```diff
@@ -77,8 +84,25 @@ def _get_reader(
...
+                            parse_nodes_uid
+                        ),
+                        SDocDocument,
+                    )
+                    assert sdoc_document.grammar is not None
+                    element = sdoc_document.grammar.elements_by_type[
+                        parse_nodes_type
+                    ]
+                    allowed_tags = [field.title for field in element.fields]
+                return SourceFileTraceabilityReader_C(parse_nodes=allowed_tags)
```
stanislaw: Let's change `parse_nodes=` to `allowed_tags` because it then communicates and mirrors the intention more directly.

strictdoc/backend/sdoc_source_code/comment_parser/marker_lexer.py
```diff
@@ -62,16 +66,23 @@ class GrammarTemplate(Template):
...
+                RELATION_MARKER_START=RELATION_MARKER_START,
+            )
             start = "(relation_marker | node_field | _NORMAL_STRING_NO_MARKER_NO_NODE | _WS)*"
         else:
+            grammar_extension = ""
             start = "(relation_marker | _NORMAL_STRING_NO_MARKER | _WS)*"
 
         grammar = GRAMMAR.substitute(
-            REGEX_NODE_NAME=REGEX_NODE_NAME,
+            GRAMMAR_EXTENSION=grammar_extension,
```
stanislaw, line 84: I am not sure if extracting this part of the grammar will not create extra confusion long-term. The argument I have against doing this is that the extracted fragment is still highly coupled to the rest of the grammar, so it is not really a standalone unit being extracted.

stanislaw (commented):
This change makes very good sense. Please proceed with implementing it.

I am only giving two comments for now, but they do not change any of the main structure that you proposed.

strictdoc/backend/sdoc_source_code/caching_reader.py
haxtibal: Ok. What would be correct naming here? Is it really "tags", or is it "fields"?

strictdoc/backend/sdoc_source_code/comment_parser/marker_lexer.py
haxtibal: I did this because, when there are zero allowed tags, what should be the grammar definition of the `node_field` symbol?

I.e., there's no more possible match. I was looking for a lark keyword to put at the right side of a BNF rule that means "no possible match", but afaikt there's no such keyword.  So the symbol should vanish completely from the grammar.  And if it vanishes, all other symbols that are only reachable through  `node_field` should also vanish. So a whole block of python t-string text shall vanish programmatically. I'm quite unfamiliar yet with t-strings, pulling the block out as separate template was the only idea I had.

What would you suggest?

strictdoc/backend/sdoc_source_code/caching_reader.py
stanislaw: `custom_tags` is ok for now. I don't have a strong opinion between custom vs allowed and tags vs fields 🤔

strictdoc/backend/sdoc_source_code/comment_parser/marker_lexer.py
stanislaw: Yesterday I didn't recognize the need in defaulting this variable to nothing when no fields are provided. Now it makes sense after your explanation.

The only thing that I would change, possibly later, is a name `GRAMMAR_EXTENSION` to make it express that this is a grammar that gets activated only if allowed fields / custom tags are provided. But I don't feel like this is super important to be addressed before merging this.

stanislaw:
> The grammar still looks for RESERVED_KEYWORDS, which are "TODO", "FIXME", and such. Now with the explicit list of tags, would you say this is still required? I don't think so, but may miss some cases.

I would say, let's keep it for the time being. Not sure enough myself.

PR #2539, opened 2025-10-30
feat(backend/sdoc_source_code): Support merging sdoc nodes with soure nodes

haxtibal:
This allows to have a (maybe incomplete) definition of a requirement anywhere in the static document tree which is then merged with content from annotated source code at runtime. Previously, StrictDoc would always have auto-generated a new node in an auto-determined document position.

Match criterion is currently the UID. This should be enhanced to MID in an upcoming change.

If the grammar element of sdoc and source code node are different, an error will be raised.

There are two use cases:
- Control the position (document, section) of a generated source node, as opposed to letting the node generator use an auto generated section.
- Support "side-car" files where parts of a requirement are defined in *.sdoc, while other parts of the same requirement are inlined with source code. This is a concept from the Linux kernel requirement template proposal.

The PR includes two related changes as groundwork. Let me know if you prefer separate pull requests.
- Bugfix to make source node file matching work on Windows. The issue existed for long, but went unnoticed since our tests didn't check field values in the rendered docs.
- Introduce GraphDatabase API to check for presence of  a particular link nodeA <-> nodeB.


strictdoc/core/file_traceability_index.py
```diff
@@ -990,3 +939,147 @@ def compare_sdocnode_by_uid(node_: SDocNode) -> str:
...
+        return source_sdoc_node
+
+    @staticmethod
+    def merge_sdoc_node_with_source_node(
+        sdoc_node: SDocNode,
+        source_node: SourceNode,
+        parent_document: SDocDocumentIF,
+        source_node_config_entry: dict[str, str],
+    ) -> None:
+        # we can only merge if grammar element definitions are compatible
```
stanislaw: Please make all comments start with a capital letter and end with a dot. 

(There is a hand-written linter script in the repo to check this but due to some false positives I haven't activated it in the CI yet.)

stanislaw (commented):
The change looks good. Just one cosmetic comment.

stanislaw:
> The change looks good. Just one cosmetic comment.

Regarding the requirements work, feel free to leave it to me, so that merging of this technical work is not blocked. We can polish the L1/L2 missing requirements separately and later.

strictdoc/core/file_traceability_index.py
```diff
@@ -990,3 +939,164 @@ def compare_sdocnode_by_uid(node_: SDocNode) -> str:
...
+        Here we link REQ and sdoc_node bidirectional.
+        """
+        for marker_ in source_node.markers:
+            if not isinstance(marker_, FunctionRangeMarker):
+                continue
+            for req_ in marker_.reqs:
+                node = traceability_index.get_node_by_uid_weak2(req_)
+                if (
+                    node
+                    not in traceability_index.graph_database.get_link_values(
```
haxtibal: I first thought (by method name) that `GraphDatabase.has_link` is there to check if a certain link `lhs <--> rhs` is present. But actually `has_link` behaves more like `has_any_link`. Would it make sense to rename that method, and introduce a new one with 3 arguments like `has_link(link_type, lhs, rhs)`?

stanislaw: Yes, please proceed with renaming to `has_any_link` and adding `has_link(link_type, lhs, rhs)`.

haxtibal:
Test fails on Windows, but works on Linux? My spontaneous idea is that this has something to do with the UID from file path construction. Maybe some drive letter or backslash has made it into the UID. Do you have a Windows machine to verify? I sadly don't... The expected UID in my test is `SRC-NODES-BASE/src/example/example.c/example_1`.

strictdoc/core/file_traceability_index.py
haxtibal: Done as separate commit within this merge request.

PR #2543, opened 2025-11-02
feat(backend/sdoc_source_code): Support remapping custom tags to sdoc fields

haxtibal:
This introduces a config option for `source_nodes` to describe how tags from an annotated source node map to *different*  `SDocNode` fields.

It's most useful in cases where one can't control the custom tag scheme at source code side, but still wants to reuse some of the custom fields for reserved StrictDoc fields with special meaning like UID or TITLE. For example, one can now use the value from "THEIR-PROPRIETARY-ID" as UID within StrictDoc, with all the special meanings of UID.


tests/integration/features/source_code_traceability/_source_nodes/40_remap/strictdoc_config.py
```diff
@@ -0,0 +1,27 @@
+from strictdoc.core.project_config import ProjectConfig
+
+
+def create_config() -> ProjectConfig:
```
haxtibal, line 4: I opted to use the new `strictdoc_config.py` config here, because with `strictdoc.toml` we run into a [toml issue](https://github.com/uiri/toml/issues/362) where nested inline tables are not pickle-able. We need to pickle when we share project configuration with child processes. I did some quick trials, but none of the potential workarounds worked for the toml case.

stanislaw: Makes sense. I would like to deprecate the TOML config long-term. It doesn't have any advantage over the newer Python config.

tests/integration/features/source_code_traceability/_source_nodes/40_remap/strictdoc_config.py
```diff
@@ -0,0 +1,27 @@
...
+    config = ProjectConfig(
+        project_title="StrictDoc Documentation",
+        project_features=[
+            "REQUIREMENT_TO_SOURCE_TRACEABILITY",
+            "SOURCE_FILE_LANGUAGE_PARSERS",
+        ],
+        source_nodes = [
+            {
+                "path": "src/",
```
stanislaw: I feel like it is a good time to adapt this dictionary of key-values to be a properly typed class. How about we make source nodes to be a `list[SourceNodesEntry]` so that the parsed ProjectConfig directly results in a list of SourceNodesEntry without an intermediate conversion from dict?

stanislaw: The new strictdoc_config.py is not yet used by users, so we are still safe and easy to make breaking changes.

strictdoc/backend/sdoc_source_code/models/source_node.py
```diff
@@ -21,5 +22,33 @@ class SourceNode:
...
+    def get_sdoc_field(
+        self,
+        field_name: str,
+        cfg_entry: SourceNodesEntry,
+    ) -> Optional[str]:
+        if field_name in cfg_entry.sdoc_to_source_map:
+            field_name = cfg_entry.sdoc_to_source_map[field_name]
+        return self.fields.get(field_name)
+
+    def sdoc_fields(self, cfg_entry: SourceNodesEntry) -> dict[str, str]:
```
stanislaw: I would prefer this to be called `get_sdoc_fields` for the symmetry with `get_sdoc_field`.

strictdoc/core/traceability_index_builder.py
```diff
@@ -881,15 +898,9 @@ def _filter_nodes(
 
     @staticmethod
     def source_node_grammar_element(
-        path_to_file: str,
-        project_config: ProjectConfig,
+        source_nodes_cfg_entry: SourceNodesEntry,
```
stanislaw: It looks like this method could depend only on the `node_type: str` without pulling the whole `source_nodes_cfg_entry: SourceNodesEntry`. If so, we could probably move it to the traceability index itself under the name `traceability_index.get_grammar_element(document_uid: str, node_type: str)`.

stanislaw (approved):

PR #2545, opened 2025-11-03
fix(export/html): Consider MARKUP = Text option in source file view

### LINKED ISSUE #2544: Bug: SourceFileViewHTMLGenerator disregards MARKUP = Text configuration
## Describe the bug

Documents can have a [MARKUP = Text](https://strictdoc.readthedocs.io/en/stable/stable/docs/strictdoc_01_user_guide.html#6.2.1.1.2-MARKUP) config option. For requirements from external sources (e.g. from source nodes) it makes sense to set this to `Text` instead of `RST`, because external content is likely to contain character sequences that would confuse an RST parser.

However, setting `MARKUP = Text` on a source nodes parent document doesn't help in that regard. StrictDoc will still  hard code `MarkupRenderer.create(markup="RST", ...) ` in `SourceFileViewHTMLGenerator.export`. In consequence, an "problems when converting RST to HTML" error is raised when StrictDoc tries to render the source code side requirements view as RST.

## To Reproduce

Create a source node parent document containing
```
[DOCUMENT]
TITLE: Requirements from Custom Source Code Tags
UID: SOURCE_NODE_DOC
OPTIONS:
  MARKUP: Text
```

Create `example.c` containing
```c
/**
 * FOO: The pointer '*ppos' does this and that.
 */
void example_1(void *ppos) {
    print("hello %p\n", ppos);
}
```

Run `strictdoc export`, it fails like

```
error: problems when converting RST to HTML: RST markup syntax error on line 24: Inline emphasis start-string without end-string.
```

## Expected behavior

One shall be able to effectively disable RST rendering for source nodes.

## Environment

Development environment on Linux using main latest.

haxtibal:
Previously, `RST` was hard-coded.

Fix it by checking on a per-node basis if `RST` or `Text` rendering shall be used. It's not possible to do this on a per-source-file basis, since requirements for one source file may come from multiple documents and the MARKUP option is attached to that documents, not to the source file.

Fixes #2544.

### DIFF
--- a/strictdoc/export/html/generators/source_file_view_generator.py
+++ b/strictdoc/export/html/generators/source_file_view_generator.py
@@ -20,6 +20,7 @@ from pygments.lexers.special import TextLexer
 from pygments.lexers.templates import HtmlDjangoLexer
 from pygments.util import ClassNotFound
 
+from strictdoc.backend.sdoc.constants import SDocMarkup
 from strictdoc.backend.sdoc_source_code.models.function_range_marker import (
     ForwardFunctionRangeMarker,
     FunctionRangeMarker,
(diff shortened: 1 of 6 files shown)


stanislaw (approved):
I think the 'get markup' part could go to the SDocNode class eventually but it is not important to do it right away.

PR #2548, opened 2025-11-05
fix(backend/sdoc_source_code): Don't auto-generate sections for merged source nodes

haxtibal:
Until now, StrictDoc unconditionally created a section for each file path component of a source node. This is however not needed in cases where the source node is merged into a static node, as the static node already has a static parent section. The auto-generated section would appear as unexpected empty reminder, so don't generate them in such a case.


stanislaw (approved):
❤️ ❤️ ❤️

PR #2549, opened 2025-11-07
feat(backend/sdoc_source_code): Support merge by MID

haxtibal:
Until now, finding a static merge candidate for source nodes relied on having a UID field set at source code side. Now we can also use MIDs for this purpose. Merge by MID gets priority over merge merge by UID.

This requires thinking about several edge cases. See test descriptions for some of them.


strictdoc/core/file_traceability_index.py
```diff
@@ -1007,6 +1043,10 @@ def merge_sdoc_node_with_source_node(
         sdoc_node_fields = source_node.get_sdoc_fields(source_node_config_entry)
         FileTraceabilityIndex.set_sdoc_node_fields(sdoc_node, sdoc_node_fields)
 
+        if "MID" in sdoc_node_fields:
```
stanislaw: Should this be inside `set_sdoc_node_fields`?

Also, it would be useful to see an integration test that reaches here. I would like to understand why do we need to do this separately, i.e., why `sdoc_node_fields.items():` iteration with `set_field_value` below does not cover setting the MID.

strictdoc/core/file_traceability_index.py
```diff
@@ -608,15 +609,50 @@ def validate_and_resolve(
...
+                if mid is not None:
+                    sdoc_node_mid = MID(mid)
+                    merge_candidate_sdoc_node = (
+                        traceability_index.get_node_by_mid_weak(sdoc_node_mid)
+                    )
+                    if merge_candidate_sdoc_node is not None:
+                        merge_candidate_sdoc_node = assert_cast(
+                            merge_candidate_sdoc_node, SDocNode
+                        )
+                        # Simplification: Either SDoc must have a static UID that is not overwritten by source node.
```
stanislaw: I am not sure I quite get what this comment is referring to. Could you rewrite it without `Either SDoc`?

strictdoc/core/file_traceability_index.py
```diff
@@ -971,6 +1002,11 @@ def create_sdoc_node_from_source_node(
         ):
             sdoc_node_fields["TITLE"] = source_node.entity_name
         FileTraceabilityIndex.set_sdoc_node_fields(sdoc_node, sdoc_node_fields)
+        if "MID" in sdoc_node_fields:
```
stanislaw: The same comment as https://github.com/strictdoc-project/strictdoc/pull/2549/files#r2507033876.

strictdoc/core/file_traceability_index.py
haxtibal: Yes probably. I pushed this MR in rather quick & dirty state to show what's next. This is one place where I was unsure. If it's in `set_field_value`, then it becomes effective not only for the create-case, but also for the merge-case. I initially thought to gracefully handle such corner cases:

```
# example.sdoc
[STATIC_NODE]
UID: NODE1
MID: 1111

# example.c
/*
 * UID: NODE1
 * MID: 22222
 */
```

where MID would 2222 would overwrite MID 1111. But I think this gets go weird for now, so I just catch it and error out, and then moving to `set_field_value` should be fine.

strictdoc/core/file_traceability_index.py
```diff
@@ -1088,6 +1124,45 @@ def connect_source_node_requirements(
...
+
+        # A merge procedure may have overwritten the MID,
+        # in which case the graph database and search index needs an update.
+        if "MID" in sdoc_node.ordered_fields_lookup != sdoc_node.reserved_mid:
+            sdoc_mid_field = sdoc_node.get_field_by_name("MID").get_text_value()
+            if sdoc_mid_field != sdoc_node.reserved_mid:
+                # TODO:
+                # If we really want to support changing the auto-assigned MID,
+                # at least the graph database and the document search index need an update (remove old MID, add new MID).
+                # I currently struggle to update the search index.
```
haxtibal, line 1142: I think I've cleaned up up a bit now. Here is my remaining problem. I try to allow and handle this case

```
# sdoc
[SRC_NODE]
UID: SREQ-1

# example.c
/*
 * UID: SREQ-1
 * MID: 12345678
 */
example_1() { }
```

meaning a MID from source code would overwrite the earlier auto-assigned MID. However, auto-assigned MID has been entered to graph and search index early (at least, maybe also to other places?) and I would need to update them. I know how to update the graph, but couldn't figure out how to consistently update the search index.

What do you think, would it be easy enough to do these updates, or should I simply not allow that edge case (exit with error)?

EDIT: For myself, I tend to not allow it. That edge case is not needed for the Linux showcase. I don't like the idea of having to modify already established graph connections while we're still in traceability construction phase. Rather, it should be possible to conceptually separate things into a "compile" and "link" phase as you already suggested. Source node parsing and merging would be part of the compile phase. At it's end we would know all nodes with some "I would like to link to..." information, but nothing is actually linked yet. And only the final link stage will add links to the graph DB and create the search index.

strictdoc/core/file_traceability_index.py
haxtibal: Moved, same as above.

strictdoc/core/file_traceability_index.py
haxtibal: The comment was only half a sentence. The place is refactored now and the comment is removed. Hope new structure and comments in itests make Intention more clear.

strictdoc/core/file_traceability_index.py
stanislaw: I need to think about your comment but something that is not clear to me even before and in general is how we want to auto-generate the UUID for both source code and sidecar when both or either of SDoc node's or source file's MID/UUID do not exist yet. Is my understanding correct that we will not have the human-readable UID at all in the Linux context?

```
# sdoc
[SRC_NODE]
 * Has no MID or UID, or MID only but the source node may not have it initially?

# example.c
/*
 * Has no MID or UID, or MID only but the SDoc document node may not have it initially?
 */
example_1() { }
```

stanislaw: On other words, it is some sort of a chicken-and-egg problem. How are we imaging the workflow of auto-generating MID/UUID between source code and sidecars?

haxtibal: > something that is not clear to me even before and in general is how we want to auto-generate the UUID for both source code and sidecar when both or either of SDoc node or source file do not exist yet

We can look at ELISA's [trace_events.c](https://github.com/elisa-tech/linux/blob/linux_requirements_wip/kernel/trace/trace_events.c#L767) annotations and their [idgen.py](https://github.com/elisa-tech/linux/blob/linux_requirements_wip/scripts/reqs/idgen.py) script.

Source code starts of like
```
/*
 * SPDX-Req-ID: [TODO: automatically generate it]
 * ...
 */
```

Then one shall call `idgen.py generate trace_events.c` to calculate `sha256sum("linux" + "trace_events.c" + instance + code)`, where instance is the text after `SPDX-Text:` in the comment, and code is the full C-function definition without comment.

The script acknowledges the problem you have mentioned

```
# TODO: since sidecar is not yet defined, this script doesn't consider the sidecar added content to the instance.
```

I see a few options:

1. For initial uuid generation, only hash over content in source code but  neglect the sdoc part (that's what idgen.py currently does). Copying the generated UUID to sdoc is a manual step. Only the second run will have the nodes merged.
2. Start off with `SPDX-Req-ID: UUID-TICKET-123`, and `MID: UUID-TICKET-123` in related sdoc. When StrictDoc sees such a preliminary UID, it will replace it with a proper calculated hash value
3. Start off with `SPDX-Req-ID: [TODO: automatically generate it]`, and `MID: tracing.c/__ftrace_event_enable_disable` in related sdoc. Let StrictDoc merge by conventional MID and replace conventional UID with proper calculated hash value
4. Use UID (manually assigned) + MID, and merge by UID.

I have no clear favorite right now from that options. Maybe we should ask Gabriele?

> Is my understanding correct that we will not have the UID at all in the Linux context?

Yes, that's also my understanding. The pilot work nowhere mentions a UID. If we wanted one, it's up to us to propose it.

stanislaw (approved):

strictdoc/core/file_traceability_index.py
stanislaw: As discussed, the user workflow for generating source and sidecar UUIDs automatically needs to be discussed further. It is not blocking the merging of this work, though, so I will resolve this comment to unblock the review.

PR #2555, opened 2025-11-13
feat(backend/sdoc_source_code): Linux Kernel Template proposal: auto-generate SPDX-Req-ID and SPDX-Req-HKey

stanislaw:


strictdoc/backend/sdoc_source_code/models/source_file_info.py
```diff
@@ -69,6 +69,14 @@ def __init__(self, g_parts: List[Any]):  # noqa: ARG002
 
         self.markers: List[RelationMarkerType] = []
 
+        self.file_bytes: bytes = b""
```
haxtibal: So we cache all source file content as permanent part of the traceability index now, right? Problems:
- Caches get outdated. What if a file changes on disk after we have read it?
- For linux, `find . -type f \( -name "*.c" -o -name "*.h" \) -print0 | du -sch --files0-from=-` reports 1,5G which will have to be pickled and have to be placed all at once in memory (per child).

Could it be an approach to turn `self.file_bytes` into a property that reads the file content temporary on demand?

strictdoc/backend/sdoc_source_code/models/source_node.py
```diff
@@ -13,16 +13,24 @@
...
 
-@dataclass
+@dataclass(eq=False)
 class SourceNode:
+    """
+    NOTE: eq=False is needed to make this dataclass support being a dictionary key.
+    """
+
     entity_name: Optional[str]
+    byte_range: ByteRange
```
haxtibal: A SourceNode can have multiple markers, which can cover multiple *non-consecutive* ranges. Further it has multiple non-consecutive fields. However, there's only a singe `byte_range` attribute here, which can cover only a *single consecutive* range. What range is it meant to cover? For example, everything from start of first marker to end of last marker?

EDIT: I think the trick is that `SourceNode` is meant to always cover a single C function comment, and the range probably means start of C-comment to end of C-comment? If so, how about renaming to `comment_byte_range`?

strictdoc/commands/manage_autouid_command.py
```diff
@@ -91,3 +109,69 @@ def execute(
                 document_meta.input_doc_full_path, "w", encoding="utf8"
             ) as output_file:
                 output_file.write(document_content)
+
+        for (
+            trace_info_
+        ) in traceability_index.get_file_traceability_index().trace_infos:
+            assert trace_info_.source_file is not None
+            # FIXME: These conditions for skipping the writes may be insufficient.
+            if (
```
haxtibal: From here on, the code IIUC is dedicated to rewrite a single source code file. Maybe refactor it into a method `ManageAutoUIDCommand.rewrite_source_file()`?

strictdoc/commands/manage_autouid_command.py
```diff
@@ -91,3 +109,69 @@ def execute(
...
+                hash_spdx_id = bytes(get_random_sha256(), encoding="utf8")
+                hash_spdx_hash = generate_code_hash(
+                    project=bytes(
+                        project_config.project_title, encoding="utf8"
+                    ),
+                    file_path=file_path,
+                    instance=bytes(instance_bytes),
+                    code=code,
+                )
+                patched_node = MarkerWriter().write(
```
haxtibal: At least SPDX-Req-ID is meant to be remapped to MID. So we have to not only update the source code with the new calculated value, but also all links in the graph database. And also all links in the search index. This is not done yet, is it?

haxtibal:
Everything I've commented here can also be addressed in upcoming PRs. Approving right now, please decide on your own what to address now or later.

strictdoc/backend/sdoc_source_code/models/source_node.py
```diff
@@ -13,16 +13,24 @@
 from strictdoc.backend.sdoc_source_code.models.range_marker import (
     RangeMarker,
 )
+from strictdoc.backend.sdoc_source_code.models.source_location import ByteRange
 from strictdoc.core.project_config import SourceNodesEntry
 
 
-@dataclass
+@dataclass(eq=False)
```
haxtibal, line 20: `eq=False` means that dictionaries will index by object identity. Copied SourceNode objects will appear as two different SourceNodes. I'm not sure this is always what we want. An alternative could be to implement `__eq__` and `__hash__` so that they target `byte_range`.

strictdoc/cli/main.py
```diff
@@ -118,6 +119,10 @@ def _main_internal(parallelizer: Parallelizer, parser: SDocArgsParser) -> None:
         # FIXME: This must be improved.
         project_config.input_paths = [manage_config.input_path]
         # FIXME: This must be improved.
+        project_config.source_root_path = str(
+            Path(manage_config.input_path).resolve()
+        )
+        # FIXME: This must be improved.
```
haxtibal: If you see a particular problem here maybe add it to the FIXME.

strictdoc/commands/manage_autouid_command.py
```diff
@@ -11,11 +13,27 @@
...
 from strictdoc.helpers.string import (
     create_safe_acronym,
 )
 
 
+def generate_code_hash(
+    *, project: bytes, file_path: bytes, instance: bytes, code: bytes
+) -> bytes:
+    """
+    "${PROJECT}${FILE_PATH}${INSTANCE}${CODE}" | sha256sum".
```
haxtibal, line 31: The comment could explain where this comes from

```
Generate hash for drift detection as suggested by Linux kernel requirements template: 
"${PROJECT}${FILE_PATH}${INSTANCE}${CODE}" | sha256sum".
```

strictdoc/core/project_config.py
```diff
@@ -463,16 +464,18 @@ def get_relevant_source_nodes_entry(
         Returns data for the first entry from source_nodes that is a parent path of path_to_file.
         If path_to_file is absolute, source node config entries are assumed to be in the source_root_path.
         """
-        if self.source_root_path is None:
-            return None
+
+        source_root_path = self.source_root_path
+        assert source_root_path is not None
+        assert os.path.exists(source_root_path), source_root_path
```
haxtibal, line 470: Since source_root_path is a `Path`, we could use [source_root_path.exists()](https://docs.python.org/3/library/pathlib.html#pathlib.Path.exists) instead of `os.path`.

tests/unit/strictdoc/backend/sdoc_source_code/test_marker_parser.py
```diff
@@ -31,6 +32,7 @@ def test_01_basic_nominal():
             line_start=1,
             line_end=1,
             comment_line_start=1,
+            byte_range=ByteRange(0, 0),
```
haxtibal: What does `ByteRange(0, 0)` mean? Shall this even be a valid range? If it is meant to indicate "no range associated", `None` would be preferable IMO.

haxtibal (approved):

strictdoc/backend/sdoc_source_code/models/source_node.py
stanislaw: Renamed to comment_byte_range.

strictdoc/backend/sdoc_source_code/models/source_file_info.py
stanislaw: Yes, this is very important. I removed the file_bytes. Instead, the manage autouid command now re-reads the files to get the file bytes. Re-reading is probably ok because this command is currently the only one that writes files back at scale.

strictdoc/commands/manage_autouid_command.py
stanislaw: Changed.

strictdoc/commands/manage_autouid_command.py
stanislaw: Yes, it is not done for now and I left a comment saying this. I still need to review your comment on how we autogenerate UUIDs for source and sidecar when both nodes do not yet have a common identifier.

strictdoc/backend/sdoc_source_code/models/source_node.py
stanislaw: I meant precisely by object identity here because we don't copy or compare the nodes in any way. I added your comment to the code with a minor modification.

strictdoc/cli/main.py
stanislaw: Changed.

strictdoc/commands/manage_autouid_command.py
stanislaw: Changed as suggested.

strictdoc/core/project_config.py
stanislaw: It is not a Path actually. I am considering if the project should switch to using Path everywhere but until now it is mostly based on os.path.

tests/unit/strictdoc/backend/sdoc_source_code/test_marker_parser.py
stanislaw: Changed to optional as suggested.

PR #2565, opened 2025-11-19
chore(.gitattributes): normalize endings to LF

stanislaw:
This should avoid any differences between how the source files are stored on Linux and Windows developer machines.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+tests/** text=auto eol=lf
+


.gitattributes
```diff
@@ -0,0 +1,2 @@
+* text=auto eol=lf
```
haxtibal: Where we currently know `eol=lf` matters is for test cases. There we want to control the exact byte wise content of test samples and expectations. So, maybe
```
/tests/** text=auto eol=lf
```
and keep the rest at default setting?

EDIT: double-star matters, to go down recursively

stanislaw: I thought that doing it the same way would be the easiest. I am not a Windows user, so I cannot guess on what they would prefer.

Do you see anything against doing it for `*`?

haxtibal: I could imagine (but don't know) that windows text editors may fight with git:
1. git converts to LF (unix) on checkout
2. windows text editor sees LF and says "no I want CRLF" and converts to CRLF
3. git thinks the text editor conversion is an intentional change and shows it in the diff

As said, I really don't know. Maybe nothing happens and it's fine.

stanislaw: I guess now that the https://github.com/strictdoc-project/strictdoc/pull/2554 is confirmed to be passing, I think we can merge this one.

PR #2566, opened 2025-11-20
feat(commands/manage_autouid): rework the generation of MID/HASH

stanislaw:
Context: This change is the next step towards generating SPDX-Req-ID and
SPDX-Req-HKey in accordance with the Linux Kernel Requirements Template proposal.

Problem:

- When auto-generating MID and HASH, we want to only rewrite these fields. All
  other fields should stay the same for the SDoc sidecar files.
- The SPDX-Req-HKey calculation algorithm shall include all fields of the
  combined requirement node (sidecar + source code). Exception: MID and hash key fields.

Solution:

- The `ManageAutoUIDCommand` has been upgraded to generate
  the `MID/SPDX-Req-ID` (SDoc and sidecar respectively) and `HASH` (SDoc only).
- The `SDocNodeFieldOrigin` field was introduced on the SDocNodeField class
  to enable new code that prevents the source node fields to be written back
  to SDoc files when the `manage autouid` command is invoked.
- The hash calculation procedure was fixed to contain calculating the hash over
  all fields of the merged node, i.e., both the SDoc and sidecar fields. The
  previous implementation was incorrect.
- `RELATION_FIELD` previously was not written by the `SDWriter` class.
  It is implemented now as part of this changeset because otherwise the diff
  expectation in the integration test highlights that this field is not present
  in the output document.

Important notes:

- The `SPDX-Req-HKey` is listed as `Sidecar` in the proposal which means that
  a source node must always have a sidecar if a hash calculation is desired.
  We could further clarify if this is what the ELISA group wants.


strictdoc/commands/manage_autouid_command.py
```diff
@@ -157,49 +160,98 @@ def _rewrite_source_file(
...
-                comment_file_bytes=file_bytes[
-                    source_node_.comment_byte_range.start : source_node_.comment_byte_range.end
-                ],
-            )
-            rewrites[source_node_] = patched_node
+            node_rewrites: Dict[str, str] = {}
+
+            # If the source node has the SPDX-REQ-ID but it is not yet a valid
+            # SHA256 identifier, create one and patch the node.
+            existing_req_id = source_node_.fields["SPDX-Req-ID"]
```
haxtibal: This logic should only look for the reserved fields `MID` and `HASH`. Not for custom tags like `SPDX-Req-ID` and `SPDX-REQ-HKey`.  If users want custom MID/HASH, they can set up a mapping in `strictdoc_config.py`:
```
SourceNodesEntry(
    ...
    sdoc_to_source_map={
        "MID": "SPDX-Req-ID",
        "HASH": "SPDX-Req-HKey",
    }
),
```
and the code should respect that mapping by using `SourceNode.get_sdoc_field`/`SourceNode.get_sdoc_fields` instead of the raw `SourceNode.fields`.

stanislaw: I have pushed a change for this in a separate commit. Please check if it looks better now.

strictdoc/commands/manage_autouid_command.py
```diff
@@ -146,7 +149,21 @@ def _rewrite_source_file(
...
+        relevant_source_node_config = (
+            project_config.get_relevant_source_nodes_entry(
+                trace_info.source_file.in_doctree_source_file_rel_path_posix
+            )
+        )
+
+        # The relevant config item should be present at this point because
+        # the FileTraceabilityIndex must have validated it during the source
+        # node graph construction.
+        assert relevant_source_node_config is not None
```
haxtibal: This fails if a file doesn't have a related entry in `ProjectConfig(source_nodes=[...])`. In linux-strictdoc, `tools/testing/selftests/devmem/devmem.c` is such a file. I think it's generally expected that not all source files are covered by source node config. No `relevant_source_node_config` means no tags, so we could probably simply skip the file?

Note: It's different if hash calculation is meant to work also for pure `*.sdoc` projects. Don't know if that is planned by you.

stanislaw: > I think it's generally expected that not all source files are covered by source node config.
You are right. I missed to take assume this properly. Now the assert is removed.

> Note: It's different if hash calculation is meant to work also for pure *.sdoc projects. Don't know if that is planned by you.

I think the answer is a clear yes but implementing this would be a separate work. I am concentrating on what Linux wants for now by going to SDoc nodes from source nodes, i.e., `sdoc_node: SDocNode = source_node_.sdoc_node`. 

**EDITED**: To support both the Linux case with the sidecar-source split and the general case without this split, the current code then has to be adjusted further to generate HASH for SDoc nodes that don't have any connection to any source files with source nodes.

haxtibal:
> > , the `ManageAutoUIDCommand` will remove `ROLE: Test` and `FUNCTION: test_devmem_access` on write-back.
> > Expectation is to not touch any attributes besides MID and HASH.
> 
> This should be addressed here: #2573. Writing back `FUNCTION` was missing but I could not reproduce that the `ROLE: Test` would be removed 🤔

I confirm `FUNCTION` is fixed now.

`ROLE: Test` is still reproducible. Will try to isolate this into a failing test case.

strictdoc/commands/manage_autouid_command.py
haxtibal: It works now for the current linux-strictdoc case. We can consider it as resolved and come back when need for hash at sdoc side becomes more evident.

strictdoc/commands/manage_autouid_command.py
haxtibal: Looks mostly good now. There are still some variable names like `hash_spdx_hash` that could be generalized simply to `hash`, I think. But not very important.

strictdoc/commands/manage_autouid_command.py
stanislaw: Yes, we handle the SDoc-level hash calculation separately.

PR #2575, opened 2025-11-25
fix(backend/sdoc_source_code): let lark scan CR to NEWLINE, not to NODE_STRING_VALUE

haxtibal:
#2554 added dedent logic, #2555 added hash generation. Tests revealed that hash results for same input were different on Windows and Linux even if CR LF was normalized to LF.

Further investigation revealed that dedent logic didn't work as expected on Windows, because CR LF got split half into the string token, and half into the newline token, which in turn confused the dedent logic. By trial it turned out that `"/[^\n\r]+/x"` (verbose regex mode, normal string => real newlines) is not the same as `"/[^\\n\\r]+/"` (normal regex mode, normal string but escaped backslash => symbolized newlines). The former wrongly lets lark scan a \r into NODE_STRING_VALUE.

Thus change to the latter pattern. Note: This fixes only the dedent issue. Hash generation still expects bit-wise identical input and tests would fail if expected hash was sampled on Linux but git converted input `\n` to `\r\n` on Windows. We need to re-add `code = code.replace(b"\r\n", b"\n")` if hash shall be agnostic to line endings.

### DIFF
--- a/tests/unit/strictdoc/backend/sdoc_source_code/test_marker_lexer.py
+++ b/tests/unit/strictdoc/backend/sdoc_source_code/test_marker_lexer.py
@@ -4,7 +4,7 @@
 
 from typing import Any, List, Optional
 
-from lark import Tree
+from lark import Token, Tree
 
 from strictdoc.backend.sdoc_source_code.comment_parser.marker_lexer import (
     MarkerLexer,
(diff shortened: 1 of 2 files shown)


tests/unit/strictdoc/backend/sdoc_source_code/test_marker_lexer.py
```diff
@@ -489,6 +489,46 @@ def test_34_node_text_starting_below() -> None:
...
+    node_fields = list(tree.find_data("node_field"))
+
+    assert_node_field(
+        node_fields[0],
+        "FIELD",
+        [
+            Token("NODE_STRING_VALUE", "value1"),
+            Token("NEWLINE", "\r\n"),
+            Token("NODE_STRING_VALUE", "value2"),
+            # The implicit \r\n => \n conversion at EOF is not nice, but doesn't hurt (yet).
```
haxtibal, line 525: Stumbled upon this during tests and tried to fix it (don't `rstrip` and convert input in `MarkerLexer.parse`, let closing comment lines `" */"` also terminate fields). However it got more involved than I hoped, so I put it aside.

stanislaw (approved):
> Thus change to the latter pattern. Note: This fixes only the dedent issue. Hash generation still expects bit-wise identical input and tests would fail if expected hash was sampled on Linux but git converted input \n to \r\n on Windows. We need to re-add code = code.replace(b"\r\n", b"\n") if hash shall be agnostic to line endings.

We didn't remove the code.replace actually, it is still there. I will try to remove the gitattributes now.

PR #2579, opened 2025-11-27
feat(commands/manage_autouid_command): update relations to point to new auto-generated MID

stanislaw:


strictdoc/core/traceability_index.py
```diff
@@ -537,6 +541,61 @@ def parent_cycle_traverse_(node_id: str) -> List[Any]:
                 datetime.datetime.today(),
             )
 
+    def update_node_mid(self, node: SDocNode, new_mid: str) -> None:
```
haxtibal, line 544: Just a question, not blocking the review: Is this something that can / should be reused when a MID is changed through server / UI?

strictdoc/core/traceability_index.py
```diff
@@ -537,6 +541,61 @@ def parent_cycle_traverse_(node_id: str) -> List[Any]:
...
+        child_nodes: OrderedSet[SDocNode] = self.graph_database.get_link_values(
+            link_type=GraphLinkType.NODE_TO_CHILD_NODES,
+            lhs_node=node,
+            edge=ALL_EDGES,
+        )
+        for child_node_ in child_nodes:
+            for relation_ in copy(child_node_.relations):
+                if (
+                    isinstance(relation_, ParentReqReference)
+                    and relation_.ref_uid == old_mid
```
haxtibal, line 574: Can we (hypothetically, very unlikely) run into a case where
- doc1 has `RELATION_FIELD: MID` and a requirement with MID "123"
- doc2 has `RELATION_FIELD: UID` and a requirement with UID "123"
- some `relation_.ref_uid` happens to mean UID "123", not MID "123?

Could we distinguish if a reference means MID "123" instead of UID "123"? If it's easy, maybe add the check here.

haxtibal (approved):
LGTM. Review comments are for curiosity, not blocking.

strictdoc/core/traceability_index.py
stanislaw: I have added the `config.relation_field != "MID":` check. I think it covers the case you described.

strictdoc/core/traceability_index.py
stanislaw: It is painful to think about this but supporting the MID/HASH autogen in the UI is its own non-trivial epic.

stanislaw: I have opened a ticket to track the UI work and your comment: https://github.com/strictdoc-project/strictdoc/issues/2580.

PR #2587, opened 2025-12-04
Add ability to render git information in metadata fields for html2pdf

mk-xw:
It would be useful for us to be able to include git info (commit, data, branch etc.) in document metadata when converting to pdf.

After digging around in the docs and source code it looks like this funcionality already existed in some capacity so I have repurposed that.

Let me know if anything needs changing!

This is a snippet from .jinja file used to test 

```
    {% if document_config.has_custom_metadata %}
    {% for key, value in document_config.get_custom_metadata() %}
    <sdoc-meta-label data-testid="document-config-metadata-label">{{key}}:</sdoc-meta-label>
    <sdoc-meta-field data-testid="document-config-metadata-field" style="min-height: 60px;">
      <sdoc-autogen>{{ value }}</sdoc-autogen>
    </sdoc-meta-field>
    {%- endif -%}
    
    {% endfor %}
```

### DIFF
--- a/strictdoc/export/html/generators/view_objects/document_screen_view_object.py
+++ b/strictdoc/export/html/generators/view_objects/document_screen_view_object.py
@@ -224,6 +224,23 @@ class DocumentScreenViewObject:
 
         return interpolate_at_pattern_lazy(self.document.config.date, resolver)
 
+    def render_metadata_value(self, metada_value:Optional[str]) -> Optional[str]:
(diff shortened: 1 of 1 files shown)


stanislaw:
To test this out, please install from the latest main branch and let me know if it works now:

```
```

I am preparing a small release and planning to deploy it by the end of this week. It will contain your contribution.

Issue #2213, opened 2025-05-03
Feature: Language-aware traceability for Rust code


fkromer:
## Description

In safety and security related contexts the usage of Rust is increasing.

## Problem

Requirements cannot be parsed from Rust files yet.

## Solution

Add parsing requirements from comments contained in Rust production code.

## Additional Information

Relates to [docs about language aware parsing of source code](https://strictdoc.readthedocs.io/en/latest/latest/docs/strictdoc_01_user_guide.html#SECTION-UG-Language-aware-parsing-of-source-code).

adfernandes:
Hi, @fkromer (and @stanislaw and @mettta)!

I'm a member of the [Safety-Critical Rust Consortium](https://github.com/rustfoundation/safety-critical-rust-consortium) and have a tiny startup using Rust for aerospace.

I've been trial using `StrictDoc` internally for a little while and... _wow_... what a nice tool!

I've made a **preliminary** tool to extract `@relation`s from Rust source code [here](https://github.com/corten-systems/sdoc-rs-relations).

Critically, it can automatically determine the context and scope of a `@relation` because Rust allows attaching [`attributes`](https://doc.rust-lang.org/reference/attributes.html) to most nodes of the Rust AST...

... and a common form of `attribute` is the ["doc-comment"](https://doc.rust-lang.org/reference/comments.html#doc-comments).

An example of using the mapped relations to generate an interactive HTML source code map is via these links:
* [relations.rs](https://corten.systems/demo/sdoc-rs-relations/relations.html)
* [relations1.rs](https://corten.systems/demo/sdoc-rs-relations/relations1.html)
* [relations2.rs](https://corten.systems/demo/sdoc-rs-relations/relations2.html)
* [relations3.rs](https://corten.systems/demo/sdoc-rs-relations/relations3.html)

**I'd appreciate your thoughts and any feedback you have on this approach.**

_(We're actually taking a similar approach to mapping object code back to source, and then source back to requirements...)_ 😄

adfernandes:
Great - thanks for taking a look!

I just wanted set of experienced “eyeballs and brains” to glance at it and give me a quick “go / no-go” to make sure I didn’t overlook something obvious.

Everything we’re doing is open-source, so as it evolves it should be good for everyone.

stanislaw:
Hi @adfernandes, thanks for sharing links for your project. It looks very close/similar to what we were going to do for supporting Rust in StrictDoc.

[...]

PR #2672, opened 2026-02-06
feat(rust): Add language aware parsing for Rust

haxtibal:
### WHAT
This adds Rust-support to StrictDoc's language aware parsing feature. It notably supports Rust doc comment semantics to automatically bind comments to language constructs. It supports line/range markers from normal comments, custom tags from doc comments and forward relations by qualified path for an opinionated set of identifiable items.

A demo of all features is available at [haxtibal.github.io/strictdoc-rust-demo](https://haxtibal.github.io/strictdoc-rust-demo/src/requirements.html).

Additional related changes are described in commit messages.

### WHY
Rust is a frequently used language in domains where also StrictDoc is used (system programming, safety critical, security critical). See related feature request #2213.

### HOW
We have two alternative proposals for doing Rust parsing:

1) Use tree-sitter-rust Python module and reimplement some of the high-level Rust doc comment semantics on our own.
2) Use syn crate (has doc comments semantics built-in), use json to transfer results to Python world.

This implements 1) and can be used to study pros and cons. It uses tree-sitter queries to match the parse tree for valid doc comment and normal comment patterns, where captures record relevant information, then python side post processing transforms captures to StrictDoc structures.

### DIFF
--- a/docs/strictdoc_21_L2_StrictDoc_Requirements.sdoc
+++ b/docs/strictdoc_21_L2_StrictDoc_Requirements.sdoc
@@ -1353,6 +1353,371 @@ Robot framework's markup has to be handled with a dedicated parser because of th
 
 [[/SECTION]]
 
+[[SECTION]]
+MID: e18dcfcb9b99459d9c7f560f35d01ff8
+TITLE: Language-aware parsing of Rust source code
+
+[REQUIREMENT]
+MID: d83cab2dcd014d328b348acc37c6f4a6
+UID: SDOC-SRS-164
+TITLE: Auto-scoped relation markers in Rust docs
+STATEMENT: >>>
+If a StrictDoc relation marker is inside a `Rust doc comment`_, the marker scope shall be set to exact the item the Rust
+language defines as target of the containing doc comment. The :code:`@relation(scope=...)` parameter shall therefore
+be optional and shall be ignored if provided.
+
+.. code-block:: rust
+   :number-lines:
+
+   /// @relation(REQ-1)
+   impl Foo {
+
+       /// @relation(REQ-2)
+       fn foo() {
+       }
+   }
+
+In the given example, the first marker shall link REQ-1 to the :code:`impl Foo` implementation,
+highlighting lines 1 to 7. The second marker shall link REQ-2 to the :code:`fn foo()` function,
+highlighting lines 4 to 6.
+
+.. _Rust doc comment: https://doc.rust-lang.org/rust-by-example/meta/doc.html#doc-comments
+<<<
+COMMENT: >>>
+Note that doc comments are syntactic sugar for `doc attribute`_ s, which shall also be supported.
+To figure detailed doc comment rules from the rust-lang specification it's sometimes useful
+to check for general attribute rules, if explicit doc comment rules are missing.
+
+.. _doc attribute: https://doc.rust-lang.org/rustdoc/write-documentation/the-doc-attribute.html
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: SDOC-SRS-34
+- TYPE: Parent
+  VALUE: SDOC-SRS-137
+
+[REQUIREMENT]
+MID: 91c7b9fd9c514327bb012639cdf1515b
+UID: SDOC-SRS-165
+TITLE: Inner doc attributes
+STATEMENT: >>>
+StrictDoc shall support relation markers in inner_ `doc attribute`_ s.
+
+.. _inner: https://doc.rust-lang.org/reference/attributes.html#r-attributes.inner
+.. _doc attribute: https://doc.rust-lang.org/rustdoc/write-documentation/the-doc-attribute.html#the-doc-attribute
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: SDOC-SRS-164
+
+[REQUIREMENT]
+MID: 002bfab5383e4cb785049c4738651f43
+UID: SDOC-SRS-166
+TITLE: Inner line docs
+STATEMENT: >>>
+StrictDoc shall support auto-scoped relation relation markers in `inner line docs`_.
+
+.. _inner line docs: https://doc.rust-lang.org/reference/comments.html#grammar-INNER_LINE_DOC
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: SDOC-SRS-164
+
+[REQUIREMENT]
+MID: d04b97d76cc848f982610035bf7d3dc2
+UID: SDOC-SRS-167
+TITLE: Inner block docs
+STATEMENT: >>>
+StrictDoc shall support auto-scoped relation relation markers in `inner block docs`_.
+
+.. _inner block docs: https://doc.rust-lang.org/reference/comments.html#grammar-INNER_BLOCK_DOC
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: SDOC-SRS-164
+
+[REQUIREMENT]
+MID: 84b351267cfc4e16b3ff2c5db654e9f0
+UID: SDOC-SRS-168
+TITLE: Outer doc attributes
+STATEMENT: >>>
+StrictDoc shall support auto-scoped relation markers in outer_ `doc attribute`_ s.
+
+.. _outer: https://doc.rust-lang.org/reference/attributes.html#r-attributes.outer
+.. _doc attribute: https://doc.rust-lang.org/rustdoc/write-documentation/the-doc-attribute.html#the-doc-attribute
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: SDOC-SRS-164
+
+[REQUIREMENT]
+MID: 171978da9f8f4f40b52fb7599caa67e6
+UID: SDOC-SRS-169
+TITLE: Outer line docs
+STATEMENT: >>>
+StrictDoc shall support auto-scoped relation markers in `outer line docs`_.
+
+.. _outer line docs: https://doc.rust-lang.org/reference/comments.html#grammar-OUTER_LINE_DOC
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: SDOC-SRS-164
+
+[REQUIREMENT]
+MID: 1a8d558fbe244621818b6242be365638
+UID: SDOC-SRS-170
+TITLE: Outer block docs
+STATEMENT: >>>
+StrictDoc shall support auto-scoped relation markers in `outer block docs`_.
+
+.. _outer block docs: https://doc.rust-lang.org/reference/comments.html#railroad-OUTER_BLOCK_DOC
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: SDOC-SRS-164
+
+[REQUIREMENT]
+MID: e1be1f38dd374f82abd9691d081f293b
+UID: SDOC-SRS-171
+TITLE: File, line and range markers
+STATEMENT: >>>
+StrictDoc shall point to lines and line ranges of Rust code by using file, line and range markers.
+These marker types shall be parsed from `regular line and block comments`_.
+
+Note: File markers may also result from the inner doc comments of the top-level module.
+
+.. _regular line and block comments: https://doc.rust-lang.org/rust-by-example/hello/comment.html#comments
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: SDOC-SRS-139
+- TYPE: Parent
+  VALUE: SDOC-SRS-124
+- TYPE: Parent
+  VALUE: SDOC-SRS-138
+
+[REQUIREMENT]
+MID: b9b2e4d377db42f5b3fb7aeabaf3f72a
+UID: SDOC-SRS-172
+TITLE: Source nodes from doc comments
+STATEMENT: >>>
+If a Rust doc comments contains custom tags, and the containing Rust source file matches an entry in the
+source node configuration, StrictDoc shall create document nodes from the custom tags and automatically
+link the created node with the Rust item targeted by the doc comment.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: SDOC-SRS-141
+
+[REQUIREMENT]
+MID: 9d7e371ad7a146e1968cd9bbb7a6790f
+UID: SDOC-SRS-173
+TITLE: Forward relations to Rust items
+STATEMENT: >>>
+StrictDoc shall support linking from a requirement in SDoc to a language item in Rust code by using a :code:`FILE`
+relation of type :code:`FUNCTION`. For example
+
+.. code-block:: strictdoc
+
+   [REQUIREMENT]
+   UID: REQ-1
+   RELATIONS:
+   - TYPE: File
+     VALUE: file.rs
+     FUNCTION: file::foo
+
+This shall work for all kinds of Rust items that have a `canonical path`_, including const, enum, fn, mod, static,
+struct, trait, type, union.
+
+.. _canonical path: https://doc.rust-lang.org/reference/paths.html#canonical-paths
+<<<
+
+[REQUIREMENT]
+MID: d01b2204fcb44882b991ca11ce567f52
+UID: SDOC-SRS-174
+TITLE: Forward relation by canonical path
+STATEMENT: >>>
+The identifier for forward relations shall be the `canonical path`_ as the rustc compiler would set it when invoked for
+a single file like
+
+.. code-block:: shell
+
+   rustc --crate-type rlib --emit=obj src/lib.rs
+
+Note: :code:`cargo build` usually result in a different canonical path, since it considers the file position relative to
+the crate root. Using a single file as root is a simplification, needed because StrictDoc is unaware of the Rust
+file and directory-based module hierarchy.
+
+.. _canonical path: https://doc.rust-lang.org/reference/paths.html#canonical-paths
+<<<
+COMMENT: >>>
+The canonical path of an item can be observed with :code:`objdump -t lib.o --demangle=rust`. For example, given this
+source file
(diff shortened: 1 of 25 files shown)


docs/strictdoc_21_L2_StrictDoc_Requirements.sdoc
```diff
@@ -1353,6 +1353,351 @@ Robot framework's markup has to be handled with a dedicated parser because of th
 
 [[/SECTION]]
 
+[[SECTION]]
+TITLE: Language-aware parsing of Rust source code
+
+[REQUIREMENT]
+UID: SDOC-SRS-RUST-DOC-COMMENT
```
stanislaw: Let's auto-generate the MID for all new nodes. All other nodes in this document have a MID.

docs/strictdoc_21_L2_StrictDoc_Requirements.sdoc
```diff
@@ -1353,6 +1353,351 @@ Robot framework's markup has to be handled with a dedicated parser because of th
...
+.. _doc attribute: https://doc.rust-lang.org/rustdoc/write-documentation/the-doc-attribute.html
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: SDOC-SRS-34
+- TYPE: Parent
+  VALUE: SDOC-SRS-137
+
+[REQUIREMENT]
+UID: SDOC-SRS-RUST-INNER-DOC-ATTR
```
stanislaw: The current convention for UID in this doc is `SDOC-SRS-<number>`. I can see how more convenient this convention could be for you but we don't have (yet) an automated generator of such word-based identifiers.

Introducing this new convention will break the consistency of the existing UIDs somewhat but see my other comment about the possible L2 vs L3 split.

docs/strictdoc_21_L2_StrictDoc_Requirements.sdoc
```diff
@@ -1353,6 +1353,351 @@ Robot framework's markup has to be handled with a dedicated parser because of th
...
+COMMENT: >>>
+Note that doc comments are syntactic sugar for `doc attribute`_ s, which shall also be supported.
+To figure detailed doc comment rules from the rust-lang specification it's sometimes useful
+to check for general attribute rules, if explicit doc comment rules are missing.
+
+.. _doc attribute: https://doc.rust-lang.org/rustdoc/write-documentation/the-doc-attribute.html
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: SDOC-SRS-34
```
stanislaw, line 1395: I am wondering if your very detailed SRS requirements are actually opening a new layer L3.

What makes me think so? It is because referencing nodes from the same spec level is usually an anti-pattern because ideally all spec levels should not be self-referential.

I encountered this problem before where some L2 requirements are more StrictDoc high-level requirements and some of them are more like children of these L2.

Do you think we could benefit from the clear L3 implementation-closest level where your Rust requirements would find a clear place, and the L2 requirement would only have a broad requirement towards Rust language-aware support (this requirement is by the way missing in your new requirements).

docs/strictdoc_21_L2_StrictDoc_Requirements.sdoc
```diff
@@ -1353,6 +1353,351 @@ Robot framework's markup has to be handled with a dedicated parser because of th
...
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: SDOC-SRS-141
+
+[REQUIREMENT]
+UID: SDOC-SRS-RUST-FWD-REL
+TITLE: Forward relations to Rust items
+STATEMENT: >>>
+It shall be possible to link from a requirement in SDoc to a language item in Rust code by using a :code:`FILE`
```
stanislaw: Let's avoid passive speech and rewrite this as `StrictDoc shall support linking ...`.

docs/strictdoc_21_L2_StrictDoc_Requirements.sdoc
```diff
@@ -1353,6 +1353,351 @@ Robot framework's markup has to be handled with a dedicated parser because of th
...
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: SDOC-SRS-RUST-DOC-COMMENT
+
+[REQUIREMENT]
+UID: SDOC-SRS-RUST-LINE-RANGE
+TITLE: File, line and range markers
+STATEMENT: >>>
+It shall be possible to point to lines of Rust code by using file, line and range markers.
```
stanislaw: Let's avoid passive speech and rewrite this as `StrictDoc shall support linking ...`.

docs/strictdoc_21_L2_StrictDoc_Requirements.sdoc
```diff
@@ -1353,6 +1353,351 @@ Robot framework's markup has to be handled with a dedicated parser because of th
...
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: SDOC-SRS-RUST-DOC-COMMENT-POS
+
+[REQUIREMENT]
+UID: SDOC-SRS-RUST-OUTER-DOC-POS
+TITLE: Allowed position for outer docs
+STATEMENT: >>>
+Consider `allowed position`_ s for outer doc comments:
```
stanislaw: This is not formulated with a `shall` statement like all other requirements.

docs/strictdoc_21_L2_StrictDoc_Requirements.sdoc
```diff
@@ -1353,6 +1353,351 @@ Robot framework's markup has to be handled with a dedicated parser because of th
...
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: SDOC-SRS-RUST-FWD-REL
+
+[REQUIREMENT]
+UID: SDOC-SRS-RUST-COLLAPSE-COMMENTS
+TITLE: Collapse doc comments
+STATEMENT: >>>
+Multiple scattered doc comments should be collapsed as done by the rustdoc_ tool when using the collapse-docs option
```
stanislaw: Let's do `should` -> `shall`.

docs/strictdoc_21_L2_StrictDoc_Requirements.sdoc
stanislaw: The advantage of L3 is that we could focus very strongly on the closest-to-code level.

The disadvantage of L3 is a need to main this level of detail for all other features but this is what we would do anyway if we had unlimited development budget.

The advantage of keeping everything in L2 is a more flat doc structure at the cost of forcing higher-level and lower-level StrictDoc requirements to live at the same level.

I am leaning towards a clear L3 but let's discuss and consider.

strictdoc/backend/sdoc/error_handling.py
```diff
@@ -323,8 +323,8 @@ def invalid_reference_type_item(
     def invalid_marker_role(
         node: SDocNode,
         marker: Union[
-            ForwardFunctionRangeMarker,
-            FunctionRangeMarker,
+            ForwardLanguageItemMarker,
+            LanguageItemMarker,
```
stanislaw, line 327: Would it be too much to ask if we could merge this renaming as a separate PR? This could reduce the scope of this PR and remove this more trivial mechanical part from it.

If it is too annoying, I could also jump in and split it myself. Just let me know.

strictdoc/backend/sdoc_source_code/models/language.py
```diff
@@ -0,0 +1,114 @@
...
+        declaration part, StrictDoc will automatically find and link
+        corresponding definition parts. Not vice versa: If definition parts are
+        documented, only the definition will be linked.
+
+        parent should be set to the SourceFileTraceabilityInfo that corresponds
+        to the file where the function is defined. Purpose is to provide meta
+        information like file name and path while the FileTraceabilityIndex is
+        not yet fully resolved.
+
+        name is an identifier that should be globally unique for a set of
```
stanislaw, line 50: It is the most detailed code comment that StrictDoc ever had 👍 

It makes me wondering if we should introduce an existing formalism for documenting function parameters? Otherwise the parameters are not easily to distinguish from the rest of the descriptive text.

E.g., `:param name: An identifier that should be globally unique for...`...

ChatGPT says there are three different conventions: 

1) Sphinx-style and PEP 257.
2) Google style.
3) NumPy style.

I guess we would pick the first one or just use colons like you did for `attributes:` below? 🤔

strictdoc/backend/sdoc_source_code/marker_parser.py
```diff
@@ -122,10 +141,12 @@ def _parse_relation_marker(
                 raise NotImplementedError
 
         assert len(relation_uid_elements) > 0
-        assert relation_scope_element is not None
 
-        assert isinstance(relation_scope_element.children[0], Token)
-        relation_scope = relation_scope_element.children[0].value
+        if relation_scope_element is not None:
```
stanislaw, line 159: Would it make sense to extract this change from this PR and write a couple of unit tests for C and Python?

Do I understand it right that if we merge this as-is, we could get an undefined behavior where a C function marker can be declared as `@relation(...)` (no scope) and then it will get its scope set to `None`?

strictdoc/backend/sdoc_source_code/reader_rust.py
```diff
@@ -0,0 +1,606 @@
...
+        input_buffer: bytes,
+        file_path: Optional[str] = None,
+    ) -> SourceFileTraceabilityInfo:
+        file_stats = SourceFileStats.create(input_buffer)
+        parse_context = ParseContext(file_path, file_stats)
+        traceability_info = SourceFileTraceabilityInfo([])
+        parser = ParserRun(
+            input_buffer, parse_context, traceability_info, self.custom_tags
+        )
+        parser()
```
stanislaw, line 292: We have never had such a pattern in StrictDoc until now. Do you strongly prefer this way of calling a parser to should we just call this function `.parse()`?

docs/strictdoc_21_L2_StrictDoc_Requirements.sdoc
```diff
@@ -1353,6 +1353,351 @@ Robot framework's markup has to be handled with a dedicated parser because of th
 
 [[/SECTION]]
 
+[[SECTION]]
```
stanislaw, line 1356: The User Guide will also need a few updates near the existing C/C++/Python/... sections.

docs/strictdoc_21_L2_StrictDoc_Requirements.sdoc
haxtibal: Basically ack. But I'm unsure what's the exact best practice to form requirement statements.

Must it always be "StrictDoc shall do this", "StrictDoc shall have that"? Meaning, when thinking of subject-predicate-object, must StrictDoc always be the subject? Or are sentences like "The identifier for forward relations shall be the canonical path" also ok? Here the subject is no longer StrictDoc, but some inner entity. And what about "Each user shall get a personal dashboard"? Here the subject is one from the external system context.

ISO 29148 provides syntax examples
```
[Condition][Subject][Action][Object][Constraint of Action]
or
[Condition][Action][Object]
```
which looks good, but content wise 'm not sure what are valid subjects/actions/objects.

strictdoc/backend/sdoc/error_handling.py
haxtibal: No problem. Which order would you prefer? 1st PR rename, 2nd PR rust support? Or vice versa? Vice versa it's very easy since I did the renaming already in a separate commit on top and can easily open an extra PR for that commit.

strictdoc/backend/sdoc_source_code/models/language.py
haxtibal: > It is the most detailed code comment that StrictDoc ever had 

I felt the need for it, since it's the entrypoint to a quite scattered functionality. A developer has to be aware of multiple related features and details that happen in distant places and stages to create the object correctly. I'm actually not sure if I understood everything correctly. Please correct or enhance comments accordingly.

> I guess we would pick the first one or just use colons like you did for attributes: below?

`attributes:` is only one colon away from reST/Sphinx syntax, so I'd vote to do `:attributes:` to have something that's commonly used. This will also become useful for distinction should we want to mix our source node syntax (`key: value`) with API documentation (`:param: description`).

strictdoc/backend/sdoc_source_code/marker_parser.py
haxtibal: > Would it make sense to extract this change from this PR

Not sure if you have seen it, it is already a [separate commit](https://github.com/strictdoc-project/strictdoc/pull/2672/changes/e597eff1fb67861cd1ade83d78a5218441cf27bb). Do you still prefer a PR? I can make one out of that commit.

haxtibal: > Do I understand it right that if we merge this as-is, we could get an undefined behavior where a C function marker can be declared as @relation(...) (no scope) and then it will get its scope set to None?

Yes, good catch. I'd suggest to define behavior so that calling parse(default_scope=None) mandates scope is set by user. If not, exit with error message "Line X: marker misses scope argument".

docs/strictdoc_21_L2_StrictDoc_Requirements.sdoc
haxtibal: > t is because referencing nodes from the same spec level is usually an anti-pattern because ideally all spec levels should not be self-referential.

Interesting. I sometimes decompose a certain requirement (at work or private stuff) that seems particularly complicated to many levels in place, maybe even 4 or 5 - while keeping the rest at L1 - and found that very helpful. I would have been scared away if I had to open many new level documents just to drill down into a lonely complicated requirement.

Why do I think this is ok? In my mind it's just a graph. Not a a set of layered documents. Where needed, the graph can get deep, and where not, the graph can remain shallow. And, there are no tiers. Everyone can know everything. Where useful, nodes can reference arbitrary other node (telling the "why"). Probably not best practice but a layman's approach. Yet it served me well.

haxtibal: > Do you think we could benefit from the clear L3 implementation-closest level where your Rust requirements would find a clear place, and the L2 requirement would only have a broad requirement towards Rust language-aware support (this requirement is by the way missing in your new requirements).

I do think it would be consistent - there's already a L1/L2 split, so a L2/L3 split is the logical next step.

Otoh, I wouldn't find it particularly useful. As said, I'm happy with traversing the graph. I find documents useful to group by topic - but not by level.

Anyway, you're the maintainer, you decide ;-)

strictdoc/backend/sdoc/error_handling.py
stanislaw: Definitely prefer renaming separately to go first. I see that that renaming already has some Rust-related changes, so you will have to edit that commit in order to cherry-pick it without errors.

strictdoc/backend/sdoc_source_code/models/language.py
stanislaw: > attributes: is only one colon away from reST/Sphinx syntax, so I'd vote to do :attributes: to have something that's commonly used. This will also become useful for distinction should we want to mix our source node syntax (key: value) with API documentation (:param: description).

Yes, let's go with reST/Sphinx syntax 👍 In the future, we could ask Codex/Claude to fix all this stuff automatically to be consistent.

strictdoc/backend/sdoc_source_code/marker_parser.py
stanislaw: > Not sure if you have seen it, it is already a [separate commit](https://github.com/strictdoc-project/strictdoc/pull/2672/changes/e597eff1fb67861cd1ade83d78a5218441cf27bb). Do you still prefer a PR? I can make one out of that commit.

A separate PR would be nice because there we could concentrate exclusively on adding those default markers for C functions as well and adding some unit tests to strengthen the default markers for non-Rust languages as well.

> Yes, good catch. I'd suggest to define behavior so that calling parse(default_scope=None) mandates scope is set by user. If not, exit with error message "Line X: marker misses scope argument".

Yes, an error message is nice to have there. The default scopes is a separate nice byproduct of your work 👍

haxtibal: Done, see #2677.

strictdoc/backend/sdoc/error_handling.py
haxtibal: Done, see #2678.

stanislaw: It is approved and merged there 👍

strictdoc/backend/sdoc_source_code/models/language.py
haxtibal: I'd prefer to not yet add Sphinx syntax in this PR, because it needs a bit more consideration. IIUC, the Sphinx parameter syntax is for API only. What I have written is mixed purpose (remember what/how/why/api) and I'm not sure how to split it up. Also, how strictly would we follow Sphinx syntax? Should it actually render with Sphinx?

stanislaw: Fine not to add for now. I agree that it is not clear how to stick with just one syntax consistently.

strictdoc/backend/sdoc_source_code/reader_rust.py
haxtibal: > Do you strongly prefer this way

No, only weakly ;-)

It's a straight implementation of a [MethodObject](https://wiki.c2.com/?MethodObject).

And what I wanted to express is that the modeled entity is actually not *the parser*. Rather it's one temporary *phase of parsing* for a certain file. On a parser instance, I should be able call run() multiple times - which is neither needed nor implemented here. In contrast, the ParserRun shall run only once, then the object is invalid. Does this make sense?

Issue #2507, opened 2025-10-14
Maybe Bug: Scope partially ignored if scope=function appears in C function comment


haxtibal:
Given a C source file

```C
void bar(void) {
}

// @relation(REQ-1, scope=file)
void foo(void) {
}
```

StrictDoc creates a source link in REQ-1 that jumps to to `foo`, highlights `foo` lines 4-6, but is entitled `"entire file"`. This looks a bit like mishmash of both "comment above function is a function marker" and "scope=file is a file marker" handling.

Expectation: Not sure, but IMO most comprehensible rule would be if `scope=` gets priority, and the marker pointed to the whole file even if it appeared in a function comment.

@adfernandes FYI, found by comparing [your example](https://github.com/strictdoc-project/strictdoc/issues/2213) to our existing C reader behavior.

PR #2677, opened 2026-02-08
feat(parsing): Allow to infer default scope for markers

haxtibal:
### WHAT
This allows language readers to provide an inferred default scope to the marker parser.

### WHY
In some situations, the explicit scope in `@relation(REQ-1, scope=function)` provided by the user would be redundant. A good example are Rust doc comments, which unambiguously determine the targeted language construct. The construct in turn determines the scope. If a language reader has the ability to infer the scope, it should do so and not urge the user to provide it explicitly. This was triggered by https://github.com/strictdoc-project/strictdoc/pull/2672.

### HOW
We slightly change lark grammar for `@relation` to:
- at least one requirement uid,
- then maybe more separated requirement uids,
- then maybe separated scope,
- then maybe separated role.

The default scope is used if the marker contained no user provided scope. We exit with error if neither default nor user provided scope were given.

### DIFF
--- a/strictdoc/backend/sdoc_source_code/marker_parser.py
+++ b/strictdoc/backend/sdoc_source_code/marker_parser.py
@@ -6,6 +6,7 @@ from typing import Dict, List, Optional, Tuple, Union
 
 from lark import ParseTree, Token, Tree
 
+from strictdoc.backend.sdoc.error_handling import StrictDocSemanticError
 from strictdoc.backend.sdoc_source_code.comment_parser.marker_lexer import (
     MarkerLexer,
 )
@@ -33,9 +34,11 @@ class MarkerParser:
         line_end: int,
         comment_line_start: int,
         comment_byte_range: Optional[ByteRange],
+        filename: Optional[str] = None,
         entity_name: Optional[str] = None,
         col_offset: int = 0,
         custom_tags: Optional[set[str]] = None,
+        default_scope: Optional[str] = None,
     ) -> SourceNode:
         """
         Parse relation markers from source file comments.
@@ -49,6 +52,16 @@ class MarkerParser:
         also be different if the comment is part of a Python or C function in
         which case the comment_line_start will be several lines below the actual
         start of the range.
+
+        filename should be given if input_string comes from a static source file.
+        It will be used to create more helpful parsing error messages.
+
+        default_scope should be provided if the caller's language-aware parser
+        can infer the scope from the semantic comment position. Think of Rust doc
+        comments for example. If given, users are allowed to omit the scope argument
+        in a relation marker. A user provided scope argument always takes preference.
+        If neither default nor a user provided value is available,
+        StrictDocSemanticError will be raised.
         """
 
         node_fields: Dict[str, str] = {}
@@ -73,8 +86,10 @@ class MarkerParser:
                     line_start,
                     line_end,
                     comment_line_start,
+                    filename,
                     entity_name,
                     col_offset,
+                    default_scope,
                 )
                 source_node.markers.extend(relation_markers)
 
@@ -102,8 +117,10 @@ class MarkerParser:
         line_start: int,
         line_end: int,
         comment_line_start: int,
+        filename: Optional[str] = None,
         entity_name: Optional[str] = None,
         col_offset: int = 0,
+        default_scope: Optional[str] = None,
     ) -> List[Union[LanguageItemMarker, RangeMarker, LineMarker]]:
         markers: List[Union[LanguageItemMarker, RangeMarker, LineMarker]] = []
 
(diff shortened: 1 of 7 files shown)


haxtibal:
There's a trade-off. This raises StrictDocSemanticError as early as possible in MarkerParser.parse and immediately crashes the app, as opposed to deferring scope validation to somewhere in general_language_marker_processors.py.
- Pro: By raising early we avoid strange helper constructs like `class InvalidScopeMarker` or `FunctionRangeMarker.scope = None` for communicating the error condition to later stages.
- Con: The reported error message lacks a file name, which would have been a good hint to users wanting to fix the error. Because, MarkerParser.parse currently doesn't know the file name.

strictdoc/backend/sdoc_source_code/marker_parser.py
```diff
@@ -208,6 +221,17 @@ def _parse_relation_marker(
                 comment_line_start + element_.meta.end_line
             )
             markers.append(line_marker)
+        elif relation_scope is None:
+            reqs = ",".join(sorted(used_uids))
+            raise StrictDocSemanticError(
+                f"@relation marker for requirements {reqs} misses scope argument.",
+                hint="Scope can only be omitted if supported by language, as e.g. with Rust doc comments.",
```
stanislaw, line 236: This naturally asks whether we would also do the same for C functions, because there we can also infer the default scope.

tests/unit/strictdoc/backend/sdoc_source_code/test_marker_parser.py
```diff
@@ -305,6 +308,72 @@ def test_30_parser_dedents_field_lines():
...
+    "input_string,default_scope,expected_type",
+    [
+        ("@relation(REQ-1)\n", "function", FunctionRangeMarker),
+        ("@relation(REQ-1, scope=line)\n", None, LineMarker),
+        ("@relation(REQ-1, scope=line)\n", "function", LineMarker),
+    ],
+    ids=[
+        "default, no user value",
+        "no default, user value",
+        "override default with user value",
```
stanislaw, line 321: Do you think we will ever have this case when a user would want to override the default scope with a different value? 

Could this create opportunities for copy-and-paste mistakes building on top of this flexibility?

strictdoc/backend/sdoc_source_code/marker_parser.py
```diff
@@ -208,6 +221,17 @@ def _parse_relation_marker(
                 comment_line_start + element_.meta.end_line
             )
             markers.append(line_marker)
+        elif relation_scope is None:
+            reqs = ",".join(sorted(used_uids))
+            raise StrictDocSemanticError(
```
stanislaw, line 234: > Con: The reported error message lacks a file name, which would have been a good hint to users wanting to fix the error. Because, MarkerParser.parse currently doesn't know the file name.

I think we should prefer raising here and adding the file name as an argument to marker_parser, trading the extra argument in exchange for immediate error reporting.

This is a kind of question when I am not 100% sure but still prefer reporting errors as early as possible over the number of arguments concern.

tests/unit/strictdoc/backend/sdoc_source_code/test_marker_parser.py
haxtibal: > Do you think we will ever have this case when a user would want to override the default scope with a different value?

In the current Rust PR, no. The rust_reader only ever sets `default_scope="function"`, and would silently ignore markers in doc comments that are not `LanguageItemMarker` where the rationale is strict specification of doc comments rust-lang.org. Overriding with `scope=function` would change nothing, and overriding with anything else would lead to marker ignore.

For other languages, like C, it could make more sense. There we don't have strict comment to item binding. We can say that a comment on top of a function *probably* means the whole function and even use this to infer a default scope, but there's nothing in the C language that mandates it. It's IMO valid to have a comment on top of a C function with a marker that says `@relation(REQ-1, scope=file)`. Younger me already filed #2507 because I expected this flexibility prior to being biased from the Rust implementation.

> Could this create opportunities for copy-and-paste mistakes building on top of this flexibility?

Yes, that's the price for it.

strictdoc/backend/sdoc_source_code/marker_parser.py
haxtibal: Yes. We already kind of have this, see #2507. As written below, I think for C it would be best to enable auto-scope, but allow to override.

Is there immediate action for this PR required?

strictdoc/backend/sdoc_source_code/marker_parser.py
haxtibal: Done (as additional parameter).

We could maybe do more refactoring here: `MarkerParser` has only static methods and no attributes. It's a class without state. And, we have `ParseContext`, which in my understanding is that missing state handled externally by callers. Couldn't we have something like this:

```py
# move ParseContext to marker_parser.py
class ParseContext:
    ...

class MarkerParser:
     def __init__(self, file_path, file_stats):
        self.context = ParseContext(file_path, file_stats)

    def context():
        return self.context
```

This would MarkerParser give state as usual in OOP. The state can be used for better error messages inside MarkerParser. And it relieves callers from having to know details about the relation between MarkerParser and ParseContext.

If you like the idea I'll try it, but in a separate PR.

stanislaw: > If you like the idea I'll try it, but in a separate PR.

I like the idea with a small remark that this we should probably not do:

> move ParseContext to marker_parser.py

Let's make MarkerParser aware of the ParseContext but not move it because inside all Readers the parse context is used directly without any relation to the marker parser.

Please do it in a separate PR.

(Let's do this in a separate PR.)

stanislaw (approved):

PR #2698, opened 2026-02-25
docs: document the custom validations feature

stanislaw:


stanislaw:
cc @haxtibal here are initial docs for the plugin/validations feature.

Issue #1321, opened 2023-10-11
Support of markdown 


villanella:
We want to edit documents with markdown syntax, and export sdoc to markdown for the integration with task tracker system.

stanislaw:
Hi @villanella, thanks for asking! The markdown support is on the roadmap, but it is of lower priority and is not something that can be done very quickly, especially if you want to use the web interface and not just edit the text files and get the static HTML export.

StrictDoc already supports the MARKUP option per-document:

```
[DOCUMENT]
TITLE: Hello world doc
OPTIONS:
  MARKUP: HTML
```

This option could be extended to also include Markdown, which would make the HTML and other exports recognize that we are dealing with Markdown, not RST or other formats. 

Also, the `RSTWriter` class could be cloned to `MarkdownWriter` and adjusted to the specifics of the Markdown format. A good and fast Markdown renderer has to be found somewhere on GitHub as a Pip package. For RST, the `docutils` library is used as the most representative for RST. Something similar has to be found for Markdown.

Finally, the static HTML export and the web interface functions have to be adjusted to be not hard-coded to RST but also support Markdown.

All these things are definitely doable, but there are work items of higher priority for me and @mettta right now.

If you have spare time to contribute this feature yourself, I could provide pointers on how this could be done but get ready to allocate one or two months of open source work to get this done, depending on whether you have existing Python experience.

villanella:
Thank you @stanislaw for development recommendations!

stanislaw:
For anyone landing here, please also see my other and similar answer about supporting AsciiDoc: https://github.com/strictdoc-project/strictdoc/issues/1337. The considerations for supporting both formats are very similar.

villanella:
@stanislaw Is it possible to support markdown with tags in yaml-frontmatter format (like in Doorstop beta) to have spec files with *.md extension instead of *.sdoc? For us it's very useful for integrations, and makes possible editing and quick viewing files in external editors or in Github/Gitlab with markdown support.

stanislaw:
You are asking something that is somewhat perpendicular to what StrictDoc is and where it is going. 

[...]

PR #2701, opened 2026-02-28
feat: add initial Markdown support

stanislaw:
**WHAT:**

Add Markdown as a first-class format: reader/writer, Markdown-to-HTML fragment rendering, discovery/export/write-back, and focused tests.

**WHY:**

There is a long-standing ticket where Markdown was requested: https://github.com/strictdoc-project/strictdoc/issues/1321.

The feedback that the StrictDoc developers have received over the years is that the SDoc format is too unconventional. Writing requirements directly in Markdown should feel more straightforward, making the requirements workflow accessible to a wider community.

**HOW:**

- This commit is the first one ever where no coding was done by human but only
L2/L3-level requirements.
- These requirements will be further refined as we proceed with supporting
Markdown in all StrictDoc features, not just Project Index and Document screens.
- Implement a Markdown backend with markdown-it-py.
- Route markup-specific writer/renderer selection by document markup.


strictdoc/export/markdown/__init__.py
```diff
@@ -0,0 +1 @@
+
```
stanislaw: Remove

strictdoc/export/html/generators/view_objects/helpers.py
```diff
@@ -67,6 +67,9 @@ def screen_should_display_file(
     ):
         return True
 
+    if file.has_extension(".md"):
```
stanislaw:
```suggestion
    if file.has_extension(".md") or file.has_extension(".markdown"):
```

pyproject.toml
```diff
@@ -58,6 +58,9 @@ dependencies = [
     # Parsing and rendering RST.
     "docutils >= 0.22.2, == 0.*",
 
+    # Parsing Markdown in CommonMark mode.
```
stanislaw:
```suggestion
    # Parsing Markdown in CommonMark mode. Converting Markdown to HTML.
```

PR #2706, opened 2026-03-03
Update layout of GCov reports

mharshe:
Rearranges the GCov report to group the Summary per file and also provide the Overall summary for the project at the top. The append order is also changed so that new test case reports are added to the end.


mharshe:
Thanks.. I need to figure out how to fix the link to code source too, but I'll create an issue if I don't solve that soon

PR #2761, opened 2026-04-04
feat(markdown): support forward relations in Markdown reader and writer

haxtibal:
WHAT: This enhances StrictDoc's markdown capabilities to write multiple parent, child and forward relations with a convention like

```md
## foo_bar_domain_map

**UID**: `REQ-001`  
**Status**: `Active`  
**Relations**:
- **Type**: `File`  
  **Path**: `drivers/foo/bar.c`  
  **Element**: `function`  
  **ID**: `foo_bar_domain_map`  
  **Hash**: `5c2e9a1b7d4f8c3a6e0b2d1f9a7c5e3b8d6f1a4c2e9b7d0f3a5c8e1b6d2f4a9`

**Statement**:
The Foo driver shall support the creation of virtual IRQs for each processor X channel combination.

It shall therefore hook into the IRQ domain framework by registering a .map file-op.

**Rationale**:
Representing as proper OS IRQs gets all the infrastructure (masking, chaining from the summary IRQ) wired up.
```

Markdown epic: #1321.

TODO:
- [ ] ongoing review and rework of draft code that resulted from manual L3 -> AI -> test iterations.


docs/strictdoc_21_l2_high_level_requirements.sdoc
```diff
@@ -1514,30 +1514,21 @@ RELATIONS:
...
 <<<
 RATIONALE: >>>
-A class is a basic building block of most programming language. Some requirements can be naturally attached to the whole classes.
-<<<
-COMMENT: >>>
-Not all programming languages support classes, e.g., C. For these languages and non-code text files, the class markers will have no effect.
+Source code elements can often be considered as the next level to trace from a requirement. For example, a low level element
+is implemented by a function, or a test case is implemented by a test function.
+
+Sometimes it's also feasible to consider the source code elements as lean replacement for prose requirement or test-case
```
stanislaw, line 1530: This second sentence is not too clear to me:

1) It sounds more like a COMMENT than a RATIONALE.
2) An example of what is referred to here would be a good addition.

docs/strictdoc_22_l3_low_level_requirements.sdoc
```diff
@@ -326,30 +326,184 @@ As of Markdown MVP, the default and preferred syntax shall be:
...
+   - Bar
+
+The syntax shall be as follows:
+
+- Only meta fields can start a conventional list structure. Bullet lists in
+  content fields are passed through as bullet lists.
+- A list reuses markdown native bullet list syntax.
+- Values can be strings or nested lists or nested dictionaries.
+
+The internal StrictDoc representation of a conventional list structure depends on the use-case (for example, a list of
```
stanislaw, line 350: Should we make this stronger and say that the currently supported use case is Relations and nothing else or did I miss something else? 🤔

docs/strictdoc_22_l3_low_level_requirements.sdoc
```diff
@@ -326,30 +326,184 @@ As of Markdown MVP, the default and preferred syntax shall be:
...
+relations will be converted to a set of Relation objects). Lists in meta field where no such use-case is defined, StrictDoc
+shall raise a StrictDocSemanticError.
+<<<
+RATIONALE: >>>
+To not restrict their native use-case when writing content, special interpretation bullet lists is limited to meta fields.
```
stanislaw, line 355: I think the main rationale here is that StrictDoc needs to reserve the list syntax for meta fields in order to support lists of dictionaries modeling Relations as the main use case. And then the current sentence `To not restrict`-sentence goes next. Also, I would rewrite it as:

```
This requirement limits its scope to bullet lists inside meta fields only, so that normal use of Markdown bullet lists in regular Markdown is not affected.
```

docs/strictdoc_22_l3_low_level_requirements.sdoc
```diff
@@ -326,30 +326,184 @@ As of Markdown MVP, the default and preferred syntax shall be:
...
+[REQUIREMENT]
+MID: 46301b38befd408b99ab1923d3deb6e5
+UID: SDOC-LLR-211
+TITLE: Dictionary types
+STATEMENT: >>>
+StrictDoc shall support dictionary-like structures in markdown files. This is an example of a dictionary with two key-value pairs:
+
+.. code-block:: markdown
+
+   **Type**: `File` \
```
stanislaw, line 367: I think we should add somewhere close to this example another example where this dictionary would be show in a realistic context of Relations. Otherwise it can be a little misleading as to where and how exactly such a dictionary is used.

strictdoc/backend/markdown/reader.py
```diff
@@ -37,13 +46,17 @@ class MarkdownHeadingNode:
 
 @dataclass
 class ParsedField:
+    """Markdown specific intermediate precursor for a SDocNodeField"""
```
stanislaw, line 49:
```suggestion
    """Markdown-specific intermediate precursor for a SDocNodeField"""
```

strictdoc/backend/markdown/reader.py
```diff
@@ -37,13 +46,17 @@ class MarkdownHeadingNode:
...
+
     name: str
     value: str
     human_name: str
 
 
 @dataclass
 class ParsedMarkdownNode:
+    """Markdown specific intermediate precursor for a SDocNode."""
```
stanislaw, line 58:
```suggestion
    """Markdown-specific intermediate precursor for a SDocNode."""
```

strictdoc/backend/markdown/reader.py
```diff
@@ -194,6 +244,14 @@ def _parse_document_root(
         including_document_reference: DocumentReference,
         file_path: Optional[str],
     ) -> Optional[str]:
+        """
+        Extend SDocDocument with data from first H1 heading.
+
+        root_heading is expected to provide the first H1 heading from the md, as previously discovered by markdown-it-py.
```
stanislaw, line 250: It looks this comment was produced with an indentation level way above the usual StrictDoc limit of 80.

stanislaw (approved):

PR #2763, opened 2026-04-06
fix(backend/sdoc): enumerate field titles across all grammar elements without duplicates

claudiiolima:
WHY:
When a document grammar defines multiple element types (e.g., `SECTION`, `REQUIREMENT`, `CUSTOM`, ...), each with its own custom fields, the table view only displayed columns from the first grammar element. Any fields unique to other element types were silently missing from the table header and rows, making the table view incomplete and misleading for documents with multi-element grammars.

### DIFF
--- a/strictdoc/backend/sdoc/models/document.py
+++ b/strictdoc/backend/sdoc/models/document.py
@@ -259,18 +259,24 @@ class SDocDocument(SDocDocumentIF):
     def enumerate_meta_field_titles(self) -> Generator[str, None, None]:
         assert self.grammar is not None
         assert self.grammar.elements is not None
-        # FIXME: currently only enumerating a single element ([0])
-        yield from self.grammar.elements[0].enumerate_meta_field_titles()
+        seen: Set[str] = set()
+        for element in self.grammar.elements:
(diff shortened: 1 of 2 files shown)


stanislaw:
Hey @claudiiolima, your change looks good. Before I merge it, just to make sure: you committed this commit using an email that is different from your GitHub account's email, this is why your GitHub avatar is not shown near the commit.

It doesn't make a difference for me but maybe you want to re-commit this change from your GitHub account instead, so that your contribution to StrictDoc is tracked more accurately (being listed as one of the contributors). I can also do it for you if you want, just let me know.

claudiiolima:
> Hey @claudiiolima, your change looks good. Before I merge it, just to make sure: you committed this commit using an email that is different from your GitHub account's email, this is why your GitHub avatar is not shown near the commit.
> 
> It doesn't make a difference for me but maybe you want to re-commit this change from your GitHub account instead, so that your contribution to StrictDoc is tracked more accurately (being listed as one of the contributors). I can also do it for you if you want, just let me know.

I will check that.

claudiiolima:
Hey @stanislaw, it was fixed now.

PR #2767, opened 2026-04-09
feat: implement document deletion functionality

xxhoeckyxx:
# Document deletion functionality

Hey,
This is Christopher from Schiederwerk GmbH. We’re interested in using your software, but we still need a few additional features.
Looking at your roadmap, I noticed the planned feature for deleting documents in the web UI. I have already implemented this and would like to submit the corresponding code as a pull request—perhaps you could incorporate it in this form.
I hope everything works as expected. Since, as a working student, I don’t yet have the same experience as some of my colleagues, I would of course appreciate any feedback.


strictdoc/server/routers/main_router.py
```diff
@@ -1369,6 +1370,63 @@ def document_tree__create_document(
...
+    @write_router.delete(
+        "/actions/document/delete_document",
+        response_class=Response,
+    )
+    def delete_document(document_mid: str) -> Response:
+        """Delete an entire SDOC document from the project.
+
+        This endpoint is intentionally simple: it removes the underlying
+        ``.sdoc`` file from disk, rebuilds the index and redirects back to the
+        project index screen. For now, it is up to the user to ensure that
```
stanislaw, line 1386: This is the reason why this feature was not implemented yet as not-so-trivial.

This `delete_document` action should ensure that there are no other documents/nodes linking to the document being deleted. While we could integrate your change as is, it could be misleading to the users because they could get corrupted project trees after this action is run without validation.

Do you think you could figure out how to add such a validation, following the example of `delete_requirement`, another router action that takes care of this validation when a single node is deleted? For deleting a document, you would need to iterate over all document tree nodes and check with every one of them if they are free from incoming links or parent/child nodes. In addition, you indeed have to check if the document that is being deleted is not included by any other document.

stanislaw:
Hey @xxhoeckyxx, thanks for contributing this. I was away from StrictDoc for some time due to other tasks but now I am back.

I see that you are merging in the latest main branch to your branch and that creates redundant commits in this PR. Please use the `git fetch origin && git rebase origin/main` workflow instead of `git merge origin/main` to make the Git history of this PR to only contain your work and no unrelated commits.

strictdoc/server/routers/main_router.py
stanislaw: Thanks for adding the validation. It seems to be working well enough for this feature to be merged. I just added a simple validation end-to-end (e2e) test.

PR #2769, opened 2026-04-09
feat(launcher): add Tk-based launcher command and wrapper script

xxhoeckyxx:
# GUI Launcher: Start/Stop Server, Exports, Logs & Project Config

Hello everyone,

this is a third proposal featuring a slightly more substantial addition 🙂

I’ve developed an additional **GUI launcher** using **tkinter** that allows users to start and stop the server via a graphical user interface.  
It also provides a file export feature, where both the **export path** and the **file format** can be selected.

## Background

At the moment, the server runs locally on every user’s machine, and not everyone feels comfortable working with the command line.  
The launcher is intended to provide a **simple and user-friendly alternative**.

## Additional Features

The GUI also allows users to:

- display the server logs at the click of a button
- edit the configuration data of individual projects directly from the UI

> **Note:**  
> This feature is particularly important for our internal use, which is why I wanted to share it with you.

As always, feedback is very welcome 🙂

Best regards,  
Christopher

### DIFF
--- a/strictdoc/features/launcher/git_action.py
+++ b/strictdoc/features/launcher/git_action.py
@@ -1,0 +1,165 @@
+import subprocess
+import threading
+from tkinter import messagebox, simpledialog
+from typing import Any
+
+
+def _ensure_git_workspace(self: Any) -> bool:
+    if not self._ensure_workspace():
+        return False
+
+    assert self.workspace_dir is not None
+    try:
+        completed = subprocess.run(
+            ["git", "rev-parse", "--is-inside-work-tree"],
+            cwd=self.workspace_dir,
+            capture_output=True,
+            text=True,
+            check=False,
+        )
+    except Exception as exc:  # noqa: BLE001
+        messagebox.showerror("Git error", str(exc))
+        return False
+
+    if completed.returncode != 0 or completed.stdout.strip().lower() != "true":
+        messagebox.showerror(
+            "Git repository missing",
+            "The selected workspace is not a Git repository.",
+        )
+        return False
+
+    return True
+
+
+def _run_git_action(
+    self: Any,
+    action_name: str,
+    git_args: list[str],
+) -> None:
+    """Run a Git command in the current workspace in the background."""
+
+    if not _ensure_git_workspace(self):
+        return
+
+    assert self.workspace_dir is not None
+    workspace_dir = self.workspace_dir
+    git_command_text = "git " + " ".join(git_args)
+    self.set_status(f"{action_name}...")
+    self._append_log(f"[{action_name.upper()}] {git_command_text}\n")
+
+    def run_git_action() -> None:
+        try:
+            completed = subprocess.run(
+                ["git", *git_args],
+                cwd=workspace_dir,
+                capture_output=True,
+                text=True,
+                check=False,
+            )
+        except Exception as exc:  # noqa: BLE001
+            error = exc
+
+            def _handle_exc(error: Exception = error) -> None:
+                self.set_status(f"{action_name} failed.")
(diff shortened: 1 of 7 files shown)


strictdoc/features/launcher/git_action.py
```diff
@@ -0,0 +1,152 @@
+import subprocess
```
stanislaw, line 1: Let's consolidate the entire Launcher codebase in the folder `strictdoc/features/launcher`. This should apply to all files from this PR that are related to this feature.

strictdoc/features/launcher/git_action.py
```diff
@@ -0,0 +1,152 @@
...
+import threading
+from tkinter import messagebox, simpledialog
+
+
+def _ensure_git_workspace(self) -> bool:
+        if not self._ensure_workspace():
```
stanislaw: You should be running `invoke lint` locally to make sure that your Python files have correct indentation. See the Dev Guide about this and other things.

strictdoc/features/launcher/git_action.py
```diff
@@ -0,0 +1,152 @@
...
+                ["git", "rev-parse", "--is-inside-work-tree"],
+                cwd=self.workspace_dir,
+                capture_output=True,
+                text=True,
+            )
+        except Exception as exc:  # noqa: BLE001
+            messagebox.showerror("Git error", str(exc))
+            return False
+
+        if completed.returncode != 0 or completed.stdout.strip().lower() != "true":
```
stanislaw: Instead of extracting this Git-related functionality to a separate file, let's keep it inside your main `launcher.py` for now. The file will be bigger but I prefer to have separate files where their interfaces are clearly separated. The issue with this file is that it seems to be about Git but in fact it is still coupled to UI-relevant things like `messagebox.showerror` etc.

tools/strictdoc_tk_launcher.py
```diff
@@ -0,0 +1,11 @@
+"""Thin wrapper to launch the Tk-based StrictDoc launcher.
```
stanislaw: Is this file actually needed?

How is it planned to be used?

stanislaw (changes requested):
Please rebase (instead of merge) this branch against the latest main branch and squash all your commits to just one. Otherwise, it is harder to review the overall changeset.

When the branch is rebase/squashed to just one commit with your actual change, I will do the next round of review. For now, I left some very obvious comments.

tools/strictdoc_tk_launcher.py
xxhoeckyxx: This file was actually intended to serve as a wrapper for a command-line argument—unfortunately, I simply forgot to remove it—the wrapper for the command-line argument is now located in “commands/launcher_command.py”

strictdoc/features/launcher/git_action.py
xxhoeckyxx: This role was ultimately not suitable for our company, so I moved it to a separate file. That way, it’s easier to comment it out or remove it. Maybe I could rename the feature to something more descriptive so it’s clear that it’s part of the user interface, or should we just remove this feature entirely? Or move all UI files to a separate folder, as you suggested in your post above. Just let me know what you prefer, and I’ll adjust it accordingly ;)

strictdoc/features/launcher/git_action.py
stanislaw: Actually, the full check is `invoke check` which includes `lint` and `test` under the hood.

xxhoeckyxx: I'm still working on the `invoke check` - hopefully I'll be able to upload it today, but I'd like to have the changes by Wednesday at the latest.

PR #2777, opened 2026-04-23
fix: gitignore now accepts escaped characters

gillesfernandez:
The filter was too restrictive, not accepting "\\" as a first character while it is an acceptable character for git ignore files. It has now been added to the list of allowed first characters.

### DIFF
--- a/strictdoc/helpers/path_filter.py
+++ b/strictdoc/helpers/path_filter.py
@@ -16,7 +16,7 @@ REGEX_MASK_VALIDATION = rf"[^(\\|\/)]{REGEX_DOUBLE_WILDCARD}"
 DISALLOWED_CHARACTERS = ("..", "{", "}", "?", "+", "!")
 
 # "~"" comes from Windows.
-ALLOWED_FIRST_CHARACTERS = ("*", "/", ".", "_", "~")
+ALLOWED_FIRST_CHARACTERS = ("*", "/", ".", "_", "~", "\\")
 
 
 def validate_mask(mask: str) -> None:


gillesfernandez:
Tested with '\\#*.*#\\' filter, it works as expected and is visible in the project tree filters as expected.

stanislaw:
@gillesfernandez thanks for contributing! Please amend your commit to contain `fix:` instead of `bugfix:`.

```
error: the commit message(s) do not match the Conventional Commits convention:

BAD: bugfix: gitignore now accepts escaped characters
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

gillesfernandez:
saw that, just fixed.

stanislaw:
> saw that, just fixed.

You still have to `commit --amend` and change the title of your commit. The linter works at the Git commit level, not at the GitHub PR title.

gillesfernandez:
Sorry, I don't seem to manage to trigger the linter checks, I even tried to do a slight change in the commit name, but it doesn't seem to trigger...

stanislaw:
> Sorry, I don't seem to manage to trigger the linter checks, I even tried to do a slight change in the commit name, but it doesn't seem to trigger...

This is your first contribution. GitHub requires first contributions to be manually approved with every push. I just approved your latest commit and the tests are now running.

gillesfernandez:
got it, makes sense. Thanks!

stanislaw:
> got it, makes sense. Thanks!

Thanks for you contribution. I am planning to release a new Pip over this weekend.

By the way, could you give an example of how a gitignore line with \ looks like?

```
# "~"" comes from Windows.
ALLOWED_FIRST_CHARACTERS = ("*", "/", ".", "_", "~", "\\")
```

gillesfernandez:
Nice, thanks!

I gave an example in my first comment. One of my tools (maybe zed) creates temporary files named #filename.example#. The associated gitignore line becomes: `\#*.*\#`

stanislaw:
> I gave an example in my first comment. One of my tools (maybe zed) creates temporary files named #filename.example#. The associated gitignore line becomes: \#*.*\#


Thanks, I was curious to see an example, and this is exactly what I was looking for :+1:

PR #2786, opened 2026-04-27
fix(tests/end2end): End2EndTestSetup: skip __pycache__ directories when diffing in/out folders

stanislaw:
WHY: Ensures that the test diffing works correctly and the __pycache__ folder does not get in the way.

HOW: Extracted from #2766.


stanislaw:
cc @thseiler

Issue #2782, opened 2026-04-27
Feature: Configurable test detection for requirement coverage tree map


chelmertz:
## Description

I would like to use `@relation()` tags in test files and have it be categorized as test coverage, even if the test does not exist within currently hardcoded `tests/` folder.

I searched the repo and think this could be the place to focus on:

```bash
rg tests/ strictdoc/features/tree_map/generator.py -A3
169:                    if "tests/" in source_file_tuple_[0]:
170-                        node_stats.child_nodes_with_links_to_test_files = 1
171-                    else:
172-                        node_stats.child_nodes_with_links_to_source_files = 1
```

## Problem

Example languages/frameworks:

- In Golang, test files are suffixed `_test.go` and found in any directory
- In [gleam](https://gleam.run/writing-gleam/#creating-a-project), test files are located in `test/` by convention

## Solution

I could see three options:

- Hardcoded like today
  - inflexible, not enough for the example languages above
- Heuristic per language/framework
  - couples strictdoc releases to language/ecosystems and their lifecycles (playing catch-up with JavaScript libraries sounds hard)
- My preferred way: configurable with a list of globs
  - Example: `test_path_patterns = ["**/*_test.go", "test/**", "tests/**"]`
  - Defaulting to previous `tests/**` if left out provides an upgrade path
  - Puts the onus + freedom on the user, suitable for many types of projects

## Additional Information

If this is interesting, I would like to have a first go at implementing it and report back here. I realize there are many features & bugs to spend your time on :) Thanks for a great project!

stanislaw:
Hey @chelmertz, yes, let's implement your `preferred way` and introduce a `tree_map_test_path_patterns` config option. Please contribute a patch.

FYI: I added this issue to be part of an epic where there is feedback from another user: https://github.com/strictdoc-project/strictdoc/issues/2793.

PR #2805, opened 2026-05-12
feat(treemap): allow configurable test dir

chelmertz:
The new optional configuration option `tree_map_test_path_pattern` allows for choosing which e.g. `@relation()` findings that should be classified as test or source, when rendered in the tree map view.

When omitted, the previous behavior of "anything within `tests/`" gets marked as a test.

For example, `tree_map_test_path_pattern = "_test\.go$"` could be helpful for go projects.

This implements https://github.com/strictdoc-project/strictdoc/issues/2782

### DIFF
--- a/strictdoc/core/project_config.py
+++ b/strictdoc/core/project_config.py
@@ -10,7 +10,7 @@ import types
 from dataclasses import dataclass, field
 from enum import Enum
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple, Pattern
 
 import toml
 
@@ -123,6 +123,7 @@ class ProjectConfig:
         exclude_source_paths: Optional[List[str]] = None,
         grammars: Optional[Dict[str, str]] = None,
         test_report_root_dict: Optional[Dict[str, str]] = None,
+        tree_map_test_path_pattern: Optional[str] = None,
         source_nodes: Optional[List[SourceNodesEntry]] = None,
         html2pdf_strict: bool = False,
         html2pdf_template: Optional[str] = None,
@@ -293,6 +294,12 @@ class ProjectConfig:
         self.test_report_root_dict: Dict[str, str] = (
             test_report_root_dict if test_report_root_dict is not None else {}
         )
+        self.tree_map_test_path_pattern: Pattern[str] = re.compile(
+            tree_map_test_path_pattern
+            if tree_map_test_path_pattern is not None
+            else r"tests/"
+        )
+
         self.source_nodes: List[SourceNodesEntry] = (
             source_nodes if source_nodes is not None else []
         )
@@ -896,6 +903,7 @@ class ProjectConfigLoader:
         include_source_paths: List[str] = []
         exclude_source_paths: List[str] = []
         test_report_root_dict: Dict[str, str] = {}
+        tree_map_test_path_pattern: Optional[str] = None
         source_nodes: List[SourceNodesEntry] = []
         html2pdf_strict: bool = False
         html2pdf_template: Optional[str] = None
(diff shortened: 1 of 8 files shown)


chelmertz:
Hi @stanislaw, sorry that I took my time to implement this :blush: 

When implementing the suggested & agreed upon "list of globs" from #2782 I got a little discouraged by all nested loops/recursion, so I went for an old and trusted regexp-approach instead (without measuring performance, so I have nothing to stand on). I looked for usage of `glob()` and didn't find much else in the codebase, so perhaps this is at least not breaking conventions.

Please let me know if the implementation/tests aren't good enough, or any other pointers you'd like me to address.

Best regards,
Calle

strictdoc/core/project_config.py
```diff
@@ -293,6 +294,12 @@ def __init__(
         self.test_report_root_dict: Dict[str, str] = (
             test_report_root_dict if test_report_root_dict is not None else {}
         )
+        self.tree_map_test_path_pattern: Pattern[str] = re.compile(
```
chelmertz, line 297: The value was passed around as an optional _string_ up until this point, I think it's fine to make it concrete as close as possible to the property assignment.

tests/integration/features/test_reports/junit_xml/ctest/06_configurable_tests_pattern/reports/tests_unit.ctest.junit.xml
```diff
@@ -0,0 +1,91 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
chelmertz, line 1: I just copied a nearby test, which is not super slim & focused at the new change..

stanislaw: Yes, copying is of course fine but let's reduce it to the bare minimum we need to exercise this new feature.

In particular, let's remove this XML file. And also I think it should be enough to only leave this test function in the test file as it is the only one you are actually testing against:

```
TEST(MyTestSuiteName, TestName)
```

(the background for this is that StrictDoc has lots of test boilerplate and I try to reduce how much boilerplate is needed and used by each test)

strictdoc/core/project_config.py
```diff
@@ -896,6 +903,7 @@ def _load_from_dictionary(
         include_source_paths: List[str] = []
         exclude_source_paths: List[str] = []
         test_report_root_dict: Dict[str, str] = {}
+        tree_map_test_path_pattern: Optional[str] = None
```
stanislaw, line 906: I am curious if there is an opportunity for reusing the already implemented filtering mechanism. Could we use the same approach that is used for `include_doc_paths` and `include_source_paths`? These filters are based on `PathFilter`.

See around these lines:

```
path_filter_includes = PathFilter(
            include_paths, positive_or_negative=True
        )
```

This way the new `tree_map_test_path_pattern` will simply follow the same configuration interface (an array of lines to whitelist) and use the same filtering mechanism.

strictdoc/core/project_config.py
stanislaw: Yes, keeping it here is exactly what I would do as well 👍

stanislaw (changes requested):
The change looks very good overall, but I left a few comments.

strictdoc/core/project_config.py
chelmertz: Will look into this!

I would like the properties of "tests are within a certain path" but also allow "this path contains both tests & implementation" to coexist (for example to cover the go convention). I wonder what inline test definitions next to implementation ([rust example](https://doc.rust-lang.org/rust-by-example/testing/unit_testing.html)) would look like, too 🤔

PR #2818, opened 2026-05-15
feat(project_config): prefer the Python config file when both are present

stanislaw:
WHY: Ensures that the Python config file take precedence compared to the deprecated TOML config file.


stanislaw:
@xxhoeckyxx as you can see here, I am integrating one of your changes separately here because it is really a separate change compared to all other PRs that you opened. This way it is much easier to review and reason about.

PR #2831, opened 2026-05-21
feat(sdoc_source_code): add known function-like Linux macros

haxtibal:
WHAT: Let the C reader recognize problematic but wanted Linux macros that look like function declarers.

WHY: Linux is one of the code bases used to showcase and evolve StrictDoc. Some of their macros confuse our tree-sitter C parser, but still shall be usable in forward relations.


stanislaw (approved):
Nice!

PR #2832, opened 2026-05-21
fix(sdoc_source_code): skip erroneous function_definitions

haxtibal:
WHAT: Ignore function_definitions where the declarer comes after a tree-sitter error node.

WHY: In such a situation the parser is already lost, and a declarer, if any, is random. Since this is sadly an expected case in non-preprocessed C-code due to macros, and since tree-sitter is good in recovering after errors, we should not simply error out, but skip the non-meaningful node.

Following Linux [kernel/softirq.c code](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/kernel/softirq.c#n125) revealed the bug:

```c
static DEFINE_PER_CPU(struct softirq_ctrl, softirq_ctrl) = {
 	.lock	= INIT_LOCAL_LOCK(softirq_ctrl.lock),
};
```

It produces a tree-sitter function_definition with nested error nodes. Worse, the spurious sub tree, which points to unrelated code far below, happens to drive reader_c into the 'identifier_node is None' branch and StrictDoc aborted with `NotImplementedError`.

### DIFF
--- a/strictdoc/backend/sdoc_source_code/reader_c.py
+++ b/strictdoc/backend/sdoc_source_code/reader_c.py
@@ -254,7 +254,7 @@ class SourceFileTraceabilityReader_C:
                         node_, "function_declarator", raise_on_error=True
                     )
                 except LookupError:
-                    # probably confused by macro, skip node to avoid processing random subtrees
+                    # Probably confused by macro, skip node to avoid processing random subtrees.
                     continue
                 # C++ reference declaration wrap the function declaration one time.
                 if function_declarator_node is None:
--- a/strictdoc/backend/sdoc_source_code/reader_c.py
+++ b/strictdoc/backend/sdoc_source_code/reader_c.py
@@ -249,9 +249,13 @@ class SourceFileTraceabilityReader_C:
             elif node_.type == "function_definition":
                 function_name = ""
 
-                function_declarator_node = ts_find_child_node_by_type(
-                    node_, "function_declarator"
-                )
+                try:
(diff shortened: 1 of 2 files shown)


haxtibal:
I believe the new `ts_find_child_node_by_type(..., raise_on_error=True)` behavior would make sense in more - if not all - call sites, but otoh it would add quite some try/expect clutter. So I only added in the one location where I actually hit the bug and wait what you say.

stanislaw:
> I believe the new `ts_find_child_node_by_type(..., raise_on_error=True)` behavior would make sense in more - if not all - call sites, but otoh it would add quite some try/expect clutter. So I only added in the one location where I actually hit the bug and wait what you say.

I would say, let's keep it in this place for now unless you know how to reproduce and write tests for these other cases. Clutter itself would be ok as long as we have it under test. I will merge this one anyway.

strictdoc/backend/sdoc_source_code/reader_c.py
```diff
@@ -249,9 +249,13 @@ def read(
...
 
-                function_declarator_node = ts_find_child_node_by_type(
-                    node_, "function_declarator"
-                )
+                try:
+                    function_declarator_node = ts_find_child_node_by_type(
+                        node_, "function_declarator", raise_on_error=True
+                    )
+                except LookupError:
+                    # probably confused by macro, skip node to avoid processing random subtrees
```
stanislaw:
```suggestion
                    # Probably confused by macro, skip node to avoid processing random subtrees.
```

PR #2838, opened 2026-05-23
test(sdoc_source_code): add unit test for UB on linux macro

haxtibal:
This unit test provides a reproducer for the bug fixed with e3b1ec33e566 to demonstrate the issue and avoid regression.


stanislaw:
As discussed and agreed, we are fixing this by switching to tree-sitter-c for now. Fixed by #2886.

PR #2845, opened 2026-05-30
fix(feat): disable some actions on autogen content

stanislaw:
WHAT: In the web GUI, make an SDoc node readonly if it contains auto-generated content. For such nodes, their action buttons and node field text fields are simply disabled and a user cannot edit them.
    
Out of scope: Drag-and-drop for auto-generated nodes is still not disabled. This will be worked on in separate PRs.


strictdoc/export/html/_static/element.css
```diff
@@ -284,6 +284,20 @@ sdoc-node-controls .action_button {
   vertical-align: top;
 }
 
+.action_button--disabled {
```
stanislaw, line 287: @mettta please double-check this.

strictdoc/export/html/_static/form.css
```diff
@@ -438,7 +438,8 @@ sdoc-autocompletable[data-field-suffix]:not(:empty)::after {
 
 /* data-field-suffix */
 
-[contenteditable=true] {
+sdoc-contenteditable,
```
stanislaw: @mettta please double-check this.

stanislaw:
@thseiler I am integrating here the 1/3 of your PR which disabled actions for nodes and node fields. It does not handle dragging and it does not handles correct rewrites of merged nodes yet.

Issue #2878, opened 2026-06-04
Bug: Markdown document edit fails validation because content is validated as RST


willenbug:
### Describe the bug

When editing a node in a Markdown document through the web UI, the field content is validated as reStructuredText instead of Markdown. Valid Markdown such as a pipe table is rejected, so the edit cannot be saved, even though the same content renders fine on the page.

The document is correctly detected as Markdown and renders as Markdown, but the edit-time validation falls back to RST regardless of the document markup.

Tested with StrictDoc 0.22.0.

### To Reproduce

1. Create a native Markdown document (`.md`) with a section whose content contains a Markdown pipe table:

   ```
   |   | Action | Verify |
   |---|--------|--------|
   | 1 | Do the thing | |
   | 2 | Do another thing | |
   ```

2. Run `strictdoc server .` and open the document. The table renders correctly.
3. Edit that section and save.
4. Saving fails with:

   ```
   RST markup syntax error on line 2: Line block ends without a blank line
   ```

The header row `|   | Action | Verify |` starts with `| `, which docutils reads as an RST line block. The separator row `|---|...` starts with `|-` (no space), which is not a valid line-block line, so docutils raises the error.

### Expected behavior

The edit saves, since the document markup is Markdown and the content is valid Markdown.

stanislaw:
Closed with the release https://github.com/strictdoc-project/strictdoc/releases/tag/0.23.0.

PR #2879, opened 2026-06-04
fix(server, UI): validate Markdown document fields as Markdown in web UI

willenbug:
When editing a node in a Markdown document through the web interface, the field content was always validated as RST, regardless of the document markup. Valid Markdown such as a pipe table was rejected, so the edit could not be saved even though the same content rendered correctly.

The edit-time field validation now selects the fragment writer based on the document's markup (Markdown for Markdown documents, RST otherwise), mirroring the logic already used by MarkupRenderer for rendering. The fix is applied to both validation sites: the requirement form object's validate() and the create/update node transform.

### DIFF
--- a/docs/strictdoc_04_release_notes.sdoc
+++ b/docs/strictdoc_04_release_notes.sdoc
@@ -81,6 +81,11 @@ editable element.
 The ``manage auto-uid`` command now resolves source root paths correctly when
 test report paths are present, and asset copying now handles Windows paths
 consistently with Linux paths.
+
+7\) Editing a node in a Markdown document through the web interface no longer
+fails when a field contains valid Markdown such as a pipe table. The edit-time
+field validation now uses the writer that matches the document's markup,
+instead of always validating the content as RST.
 <<<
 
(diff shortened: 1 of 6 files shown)


willenbug:
Hi there!

I noticed this bug when using StrictDoc: https://github.com/strictdoc-project/strictdoc/issues/2878

This PR is meant to correct that bug.

stanislaw:
Thanks for contributing! Your changeset was almost correct, I just fixed it slightly to handle the case when editing a document included to another document.

FYI: I am not sure if it matters to you, but it looks like you are committing from an email address that is different from the email address associated with your GitHub account.

One consequence is that you will not be listed as a StrictDoc contributor because GitHub does not seem to recognize commits made from external email addresses. See here:

https://github.com/strictdoc-project/strictdoc/graphs/contributors

Maybe this is not too important to you, but I thought I would mention it.

willenbug:
Thank you very much! This is a lovely project.

And thank you for the heads up about the email!

PR #2984, opened 2026-06-30
feat(server): auto-rebuild and live-reload on file changes (--watch)

### LINKED ISSUE #2840: Feature: Auto-rebuild and live-reload on file changes for StrictDoc server
## Description

Add an automatic change-detection and live-reloading feature to the StrictDoc server. When documentation files are modified, the server should automatically rebuild the HTML output on-the-fly and display any potential compilation or structural errors.

## Problem

Currently, when editing documentation while running the StrictDoc server, changes to the document structure are not reflected in real-time. Users must manually rebuild the documentation or restart the server to see updates or to check for syntax errors. This introduces friction and slows down the documentation authoring workflow.

## Solution

Implement a file-watching mechanism within the strictdoc server command.

- **On success:** Automatically trigger a fast rebuild of the served HTML whenever changes to .sdoc (or related configuration/grammar) files are detected.
- **On error:** Capture any structural or syntax errors during the auto-rebuild and alert the user (e.g., via a terminal notification, browser overlay, etc).

## Additional Information

A similar on-the-fly rebuilding and error-alerting feature is successfully implemented in Sara (https://github.com/cledouarec/sara), another documentation tool.

tansaku:
## Summary

Adds an opt-in `strictdoc server --watch` mode that watches the project's document files and live-reloads connected browsers when documents change on disk. Closes #2840.

When a watched document changes, the server rebuilds the in-memory traceability index and notifies browsers over the existing websocket: a successful rebuild triggers a reload (the page regenerates lazily), while a build error is shown as a browser overlay and printed to the terminal, with the server staying up.

## Notes

- Opt-in via `--watch` (default off), so the existing in-UI editing flow is unaffected.
- Change detection is by file **content hash**, not mtime. StrictDoc rewrites source files with byte-identical content during generation, and reacting to those rewrites caused a reload loop. Hidden directories (e.g. `.git`, virtualenvs) and the output directory are not watched.
- `watchdog` is promoted from a dev-only to a runtime dependency.
- Reuses the existing websocket endpoint and client (`websocket.jinja.html`).

## Tests

- New unit tests in `tests/unit/strictdoc/server/test_document_watcher.py` covering extension / hidden-dir / output filtering and content-hash change detection (an identical rewrite does not reload; a real change does).
- `ruff`, strict `mypy`, `test-unit` (579) and `test-unit-server` (7) pass locally, along with the integration suite.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/docs/strictdoc_01_user_guide.sdoc
+++ b/docs/strictdoc_01_user_guide.sdoc
@@ -460,6 +460,28 @@ StrictDoc uses ``127.0.0.1`` as a default host and ``5111`` as a default port. W
     See [LINK: SDOC_UG_LIMIT_WEB] for an overview of the existing limitations.
 <<<
 
+[[SECTION]]
+MID: 3556cc0790954c0dadab5da0c80cdfcc
+TITLE: Live reload (--watch)
+
+[TEXT]
+MID: f48a8731afa84849af3c06f302d24c98
+STATEMENT: >>>
+The ``--watch`` option enables automatic detection of file changes and live reload of all connected browsers:
+
+.. code:: text
+
+    strictdoc server --watch .
+
+When ``--watch`` is active, StrictDoc monitors the input document files for changes. If a file is modified by an external tool (such as a text editor or a version control operation), StrictDoc rebuilds the documentation tree and sends a reload signal to all connected browser tabs.
+
+.. note::
+
+    Live reload is triggered only by external changes to the watched files. Edits made through the StrictDoc web interface itself do not trigger a reload, since the server already handles those changes internally.
+<<<
+
+[[/SECTION]]
+
 [[/SECTION]]
 
 [[SECTION]]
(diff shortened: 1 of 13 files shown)


stanislaw:
@tansaku thanks for contributing! Please give me a few days to consider this carefully. I just pushed my commit on top of your branch but it is not yet doing a good enough job to ensure that the editing of content in UI is not affected by this live-reload feature.

cc @larsjohansson-saferadar

stanislaw:
@tansaku @larsjohansson-saferadar

This has been merged and can be installed from the main branch for testing:

```sh
```

I am still concerned that there may be edge cases where editing something in the UI triggers the server to reload unintentionally. The fix that I pushed in my second commit should prevent this from happening, but let's test it some more to identify whether anything was missed.

larsjohansson-saferadar:
This is very nice, for my basic editing it works nicely to see the updates happen in the browser and even just seeing the result of the automatic re-parsing in the terminal in VSCode will be a useful workflow for me. Thanks so much!

The only edge case I've discovered so far is when I am being intentionally mean. I have a test document:

```[DOCUMENT]
TITLE: The document

[[SECTION]]
TITLE: Introduction

[[/SECTION]]
```

If I add an extra blank line after the title, I get the error message both in the terminal and in the browser, as expected:
image

If I fix it in VSCode, then the error clears, as expected.

But if I instead sneak behind the message in the browser and edit the title of the document (or make no changes, just re-save it)
image
then the line disappears in the VSCode editor but the message in the browser stays, despite there no longer being anything wrong with the source file. If I refresh the page, the message clears, but this also happens even if I just refresh without having fixed the problem.

PR #2996, opened 2026-07-03
fix(markdown): stop duplicating document UID in the HTML metadata panel

haxtibal:
What: Prevent rendering document UID twice (1 from config, 1 from metadata). #2995 missed to consider this.

Why: The markdown reader uses custom metadata to remember field order for write-back by storing a copy of config fields there. The renderer doesn't know about this trick and just displays config along with metadata. If we want to prevent rendered duplicates we need to filter somewhere.

### DIFF
--- a/strictdoc/backend/sdoc/models/document_config.py
+++ b/strictdoc/backend/sdoc/models/document_config.py
@@ -184,20 +184,28 @@ class DocumentConfig:
         )
 
     def has_custom_metadata(self) -> bool:
-        return (
-            self.custom_metadata is not None
-            and self.custom_metadata.entries is not None
-            and len(self.custom_metadata.entries) > 0
-        )
+        return len(self.get_custom_metadata()) > 0
 
     def get_custom_metadata(self) -> List[Tuple[str, str]]:
         if (
             self.custom_metadata is not None
             and self.custom_metadata.entries is not None
         ):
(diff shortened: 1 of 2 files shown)


strictdoc/backend/sdoc/models/document_config.py
```diff
@@ -184,20 +184,28 @@ def has_meta(self) -> bool:
...
+                    ("DATE", self.date),
+                    ("CLASSIFICATION", self.classification),
+                )
+                if value is not None and len(value) > 0
+            }
             return [
                 (entry.key, entry.value)
                 for entry in self.custom_metadata.entries
-                if entry.key is not None and entry.value is not None
+                if entry.key is not None
```
haxtibal, line 207: I'm not a fan of this filtering-after-the-fact approach and would prefer a SemanticError when someone tries to insert custom metadata that already exists in config (because, what shall it mean if a user provides one of the reserved config fields in their metadata?).

However, error-on-insert doesn't work as long as the markdown reader relies on custom meta data for order tracking. I have no quick idea around.

stanislaw: Ok for now. Let's discuss the order tracking approach separately. Maybe we could parse in any order but always write back in the fixed order.

stanislaw (approved):

PR #2997, opened 2026-07-03
feat(source-nodes): automatically link test results to test case source nodes

haxtibal:
What: automatically link an auto-generated TEST_RESULT node to the auto-generated document node of a source node.

Why: We can create nicer traceability:

```
                             (actual test implementation,
                                      e.g. robot code)
                              +----[LanguageItem]
                              |
  [REQ] <--------------> [TEST_CASE] <--------------> [TEST_RESULT]
        (relation from   (from source  (new, this change)
         merge with side   node)
         car stub)
```

How: the automatic linking matches as follows
1. for a given auto-generated TEST_RESULT node
2. find a LanguageItem where LanguageItem.name equals TEST_RESULT.TEST_FUNCTION
3. find the SourceNode that was parsed from this LanguageItem, if one exists and was already merged into a document node
4. link that SourceNode's document node and the TEST_RESULT node

### DIFF
--- a/strictdoc/core/file_traceability_index.py
+++ b/strictdoc/core/file_traceability_index.py
@@ -521,6 +521,25 @@ class FileTraceabilityIndex:
                                 rhs_node=forward_requirement_,
                                 edge="IsSatisfiedBy",
                             )
+
+                    for source_node_ in function.parent.source_nodes:
+                        if (
+                            source_node_.function is function
+                            and source_node_.sdoc_node is not None
+                        ):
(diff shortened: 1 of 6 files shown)


haxtibal:
IIRC, the source nodes and test report feature was originally requested by people who wanted it for some gtest use case. The feature I'm proposing here seems natural to me in such a context (just that it's robot framework in my case), so I wonder if the gtest people may already have implemented it and I just overlooked it?

PR #3002, opened 2026-07-04
feat(source_node): Merge source nodes with automatic TITLE as last option

haxtibal:
What: When a sidecar stub node is merged with a source node, and neither stub nor source node kv-pairs contain a title, use the entity name suggested by the individual language parsers as TITLE field.

Why: The logic was already present in the create-case and is now reflected in the merge-case. To see why it's useful: Imagine a test suite with 100 robot tests (where all test cases already have a nice title from robot DSL syntax). Imagine a side-car stub document with 100 stubs for user-defined section grouping and user-defined parent-child relations of the test cases. In this case we don't want to copy/duplicate the title over from robot DSL to StrictDoc stubs. Copies are additional work and would drift away soon.


haxtibal:
Note: This copies a commit from #2998, since it depends on it. The copy can be removed when 2998 is merged.

stanislaw (approved):

PR #3022, opened 2026-07-08
feat(query): add quantified list membership queries

larsjohansson-saferadar:
Add any/all/none query expressions for matching quoted string lists against fields.

Quantified matching keeps scalar in/not in behavior unchanged. For node["TAGS"], the new list forms match exact comma-space-separated tag values instead of substrings.

Includes parser, evaluator, documentation, and --filter-nodes integration coverage.

### DIFF
--- a/developer/tasks/20260708_query_list_membership/task.md
+++ b/developer/tasks/20260708_query_list_membership/task.md
@@ -1,0 +1,74 @@
+# Add quantified list membership expressions to the query language
+
+## WHAT
+
+Adds support for quantified list membership expressions in the StrictDoc query language:
+
+Example:
+```
+any(["tag_a", "tag_b"]) in node["TAGS"]
+all(["tag_a", "tag_b"]) in node["TAGS"]
+none(["excluded_a", "excluded_b"]) in node["TAGS"]
+```
+
+The existing scalar membership expression remains supported:
+```
+"System" in node["TITLE"]
+```
+
+## WHY
+
+StrictDoc already supports substring-style membership checks such as `"System" in node["TITLE"]`. When filtering nodes by tags or other fields, users currently have to repeat this expression with explicit boolean operators:
+
+```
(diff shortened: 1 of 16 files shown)


developer/tasks/20260708_query_list_membership/task.md
```diff
@@ -0,0 +1,74 @@
+# Add quantified list membership expressions to the query language
+
+## WHAT
+
+Adds support for quantified list membership expressions in the StrictDoc query language:
+
+Example:
+```
+any ["tag_a", "tag_b"] in node["TAGS"]
```
stanislaw: We have been using a Python-like syntax until now, so let's make it look like the usual Python `any()` syntax, i.e.,

```
any(["tag_a", "tag_b"]) in node["TAGS"]
```

Please do the same for `all` and `none` as well.

strictdoc/features/project_index/assets/project_tree.css
```diff
@@ -170,7 +170,7 @@
 .dashboard {
```
stanislaw, line 170: It looks like something went wrong with rebasing from the latest main branch. This diff is related to the change by @mettta that I merged earlier today.

Run something like `git fetch origin && git rebase origin/main` and this part of the diff should go away (in theory). If it doesn't, you will have to remove these unrelated changes manually from your diff and force push this branch.

stanislaw (changes requested):

strictdoc/features/project_index/assets/project_tree.css
larsjohansson-saferadar: I think I got rid if this.

developer/tasks/20260708_query_list_membership/task.md
larsjohansson-saferadar: Updated according to proposal.

stanislaw: Looks good!

stanislaw (approved):

PR #3042, opened 2026-07-14
feat(tests/screencast): add Python/Playwright screencast pipeline + hello_world demo scenario

mettta:
- Adds a self-contained demo-video pipeline in Python/Playwright: each scenario is a dual-purpose pytest test (`invoke test-screencast`) that also (re)generates its `.webm` via `--record-video`, backed by a real live `SDocTestServer` instance and a small Playwright Page Object layer (`tests/screencast/helpers/`). 
- Adds the `hello_world` scenario: a real `strictdoc new` run (shown in a terminal-style scene) → browsing the generated project in the web UI → following a Requirement's Parent/Child relation links across documents → editing an existing Requirement to add a Rationale → an editor-style scene revealing the real resulting `.sdoc` diff, syntax-highlighted with StrictDoc's own `StrictDocLexer`.
- Adds demo-quality infrastructure reused by every scene: `Pointer` (a fake cursor + click-target highlight, since Playwright renders no real cursor), pacing helpers for deliberate pauses, terminal/editor/IDE-typing HTML playgrounds, and web-optimized video export.
- Adds `invoke screencast-server` for manually previewing a scenario in a real browser — defaults to a disposable, read-only copy of the project (nothing persists, the shared fixture is never touched); `--edit` opts into the real, persistent files when intentionally curating them.
- Adds `README.md` documenting the whole mechanism: commands, pacing/cursor knobs, scene-editing, the read-only/`--edit` manual server, syntax highlighting.


strictdoc/export/html/_static/content.css
```diff
@@ -85,7 +85,7 @@
 [data-viewtype="document"] .content {
   display: block;
   max-width: 900px;
-  margin-bottom: 300px;
+  /* margin-bottom: 300px; */
```
stanislaw, line 88: Is this intentional and permanent? Maybe add a comment explaining why it is here in the commented out state?

PR #3048, opened 2026-07-18
Deep markdown headings

### LINKED ISSUE #3046: Feature: support Markdown headings deeper than six levels (opt-in)
## Description

Large documents will likely have headings deeper than six levels (see #3045 ), even though this is technically not supported per specification.

An opt-in, experimental project feature (`MARKDOWN_DEEP_HEADINGS`) lets the Markdown format represent documents nested deeper than six levels, plus an accessibility improvement for how the HTML export renders such deep sections.

## Problem

The `.sdoc` grammar nests sections to arbitrary depth, but the Markdown format hits a wall at six levels: CommonMark does not recognize `#######` as a heading. The current behavior fails silently as the Markdown writer happily emits 7+ hashes for a deep document, and the CommonMark-based reader then cannot parse its own output back. In my test, the round-trip died with a "duplicate field names" error, because the orphaned heading lines were swallowed into the previous node's body.

I looked at a few existing implementations to see how they handled this scenario: as already discussed, CommonMark/`markdown-it` treat `#######` as paragraph text; Pandoc's own Markdown flavor accepts unlimited hashes as an extension; `docutils`/Sphinx (RST allows arbitrary depth) clamp HTML output at `<h6>` but add `aria-level` with the true depth.

## Solution

- A new experimental project feature, `MARKDOWN_DEEP_HEADINGS`, off by default.
- If `MARKDOWN_DEEP_HEADINGS` is enabled: the reader accepts 7+ hash headings
- If `MARKDOWN_DEEP_HEADINGS` is disabled: the writer now raises a clear error naming the flag for a too-deep document, instead of silently producing unreadable output...I'd argue that part is a bug fix regardless.
- HTML export: headings clamped at `<h6>` get `aria-level` with the true depth (the docutils/Sphinx convention), so screen readers announce the correct level. Levels six and shallower are unchanged.
- A user guide subsection documenting the feature and its trade-off.

martinezlc99:
Resolves #3046

### DIFF
--- a/docs/strictdoc_01_user_guide.sdoc
+++ b/docs/strictdoc_01_user_guide.sdoc
@@ -4985,6 +4985,58 @@ document-level prefix for all requirements within that section.
 
 [[/SECTION]]
 
+[[SECTION]]
+MID: 3f33a1441f8140ba886ff91371197503
+UID: SECTION-UG-Markdown-deep-headings
+TITLE: Deep headings (more than six levels)
+
+[TEXT]
+MID: 299c3572be2d4e5c9d06b2ce327d377b
+STATEMENT: >>>
+CommonMark recognizes headings with at most six ``#`` characters, so a
+Markdown document can normally nest sections only five levels below the
+document title. SDoc documents, however, support arbitrary nesting depth.
+
+To allow deeply nested documents to round-trip through the Markdown format,
+StrictDoc deviates from CommonMark on this point: the Markdown reader accepts
+headings with seven or more ``#`` characters, following the precedent of
+Pandoc's Markdown reader, and the Markdown writer emits such headings for
+nodes nested deeper than six levels.
+
+**Example of a requirement nested seven levels below the document title:**
+
+.. code-block:: markdown
+
+    # Deep headings example
+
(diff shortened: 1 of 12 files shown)


stanislaw (changes requested):
@martinezlc99 the PR looks great.

The only comment: Let's make the `>6-level headers` the default behavior. No need to maintain the feature toggle `MARKDOWN_DEEP_HEADINGS`. 

But we should keep your note in the user guide, explaining that we deviate from the standard Markdown behavior.

martinezlc99:
> @martinezlc99 the PR looks great.
> 
> The only comment: Let's make the `>6-level headers` the default behavior. No need to maintain the feature toggle `MARKDOWN_DEEP_HEADINGS`.
> 
> But we should keep your note in the user guide, explaining that we deviate from the standard Markdown behavior.

Hi @stanislaw - been on work travel this week but will incorporate your suggestion in the next day or so.

Thank you for looking it over!

stanislaw (approved):

stanislaw:
cc @haxtibal

PR #3049, opened 2026-07-18
[DISCARDED] Chunked lazy loading

### LINKED ISSUE #3045: Feature: chunked lazy loading for large documents in the web GUI
## Description

Lazily load large documents in chunks in the web GUI, so that opening a document with thousands of nodes stays fast, plus two browser-side rendering cost reductions that help regardless of document size.

## Problem

I'm considering using StrictDoc for a requirements set that will nominally exceed 10k+ requirements. The web GUI struggles once a single document reaches thousands of nodes: the initial render takes a long time and browsing stays sluggish afterwards. 

Root causes:

1. Every node instantiated its own copy-to-clipboard Stimulus controller and listener set, and MathJax/Mermaid processing blocked first draw.
2. The document screen renders the entire document up front, no matter its size.

## Solution

1. One delegated document-level listener for all copy-to-clipboard buttons, and MathJax typesetting / Mermaid rendering
   deferred to browser idle time instead of blocking first draw.
2. Documents with more than `chunked_documents_threshold` nodes (default `200`; `0` disables the feature) render the first chunk inline and load the remaining content as lazy turbo-frames while scrolling.  A cursor-based fragment endpoint (`GET /fragments/document/{document_mid}/chunk`) implements the lazy load. TOC clicks, URL deep-links, and back/forward navigation eagerly load the chunk containing their target anchor. Small documents, non-chunked rendering, and the static HTML export are unaffected.

## Additional Information

Future PRs:

1. TOC scroll-highlighting only tracks the loaded portion of the document.
2. Ctrl+F cannot find text in chunks that have not loaded yet and a failed fragment load does not yet offer a retry UI.
3. Address static html exports for the same issue

martinezlc99:
Resolves #3045

### DIFF
--- a/docs/strictdoc_01_user_guide.sdoc
+++ b/docs/strictdoc_01_user_guide.sdoc
@@ -4430,6 +4430,40 @@ Use the ``[server]`` section to configure the host and port as follows.
 
 [[/SECTION]]
 
+[[SECTION]]
+MID: 6ebd68891d4a4d518093415cbd6f5f2e
+UID: SECTION-UG-Chunked-documents-threshold
+TITLE: Chunked documents threshold
+
+[TEXT]
+MID: ff77928f09094995acb61af071157431
+STATEMENT: >>>
+When a document contains many nodes, the StrictDoc server can render the
+document in lazily loaded chunks to speed up the initial page load. The
+``chunked_documents_threshold`` option controls the threshold number of nodes:
+the chunked rendering is activated for documents that contain more than this
+number of nodes. The default value is ``200``. Setting the option to ``0``
+disables the chunked rendering completely. Threshold values below ``100`` also
+become the chunk size: the documents are loaded in chunks of ``min(threshold,
+100)`` nodes.
+
+Use the ``[server]`` section of ``strictdoc.toml`` to configure this option.
+
+.. code:: python
+
+    from strictdoc.core.project_config import ProjectConfig
+
+
+    def create_config() -> ProjectConfig:
+        config = ProjectConfig(
+            project_title="Test project",
+            chunked_documents_threshold=100,
+        )
+        return config
+<<<
+
+[[/SECTION]]
+
 [[/SECTION]]
 
 [[/SECTION]]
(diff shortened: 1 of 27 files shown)


docs/strictdoc_01_user_guide.sdoc
```diff
@@ -4430,6 +4430,40 @@ Use the ``[server]`` section to configure the host and port as follows.
 
 [[/SECTION]]
 
+[[SECTION]]
```
stanislaw, line 4433: @martinezlc99, I have tried testing this branch against StrictDoc's own documentation. I set the threshold to a small value as follows:

```
    DEFAULT_CHUNKED_DOCUMENTS_THRESHOLD = 5
```

When I reload the User Guide document and I start scrolling slowly, I start seeing empty gaps starting from the section:

Image

I gave it to ChatGPT quickly and it produced the following change that seems to fix it however I didn't spend enough time with your PR to fully understand the mechanics. It may or not be useful for the actual fix needed here: https://github.com/strictdoc-project/strictdoc/pull/new/stanislav/testing_lazy_loading.

docs/strictdoc_01_user_guide.sdoc
```diff
@@ -4430,6 +4430,40 @@ Use the ``[server]`` section to configure the host and port as follows.
...
+MID: 6ebd68891d4a4d518093415cbd6f5f2e
```
stanislaw, line 4434: I am running out of time today to integrate @mettta's changes from https://github.com/strictdoc-project/strictdoc/pull/3066. She will take a look at your branch tomorrow and explore the minimal changes needed to make both PRs work well together.

docs/strictdoc_01_user_guide.sdoc
mettta: I added 2 follow-up commits to the `stanislav/testing_lazy_loading` branch:

- refining the lazy chunk preload wiring; see the commit message for the technical details;
- adding a minimal visual loading indication inside unloaded chunk frames.

stanislaw:
Closing this in favor of https://github.com/strictdoc-project/strictdoc/pull/3071 which we have just merged.

PR #3054, opened 2026-07-22
feat(markdown): map Version/Date/Classification to config

haxtibal:
What: Support `Version`, `Date` and `Classification` document config fields for markdown.

Why: Get closer to sdoc feature parity.

How: Recognize the keywords and store values in `DocumentConfig`.

### DIFF
--- a/docs/strictdoc_04_release_notes.sdoc
+++ b/docs/strictdoc_04_release_notes.sdoc
@@ -50,6 +50,20 @@ releases:
 <<<
 
 [[SECTION]]
+MID: a7ffb392d48244969318ec115e4695bd
+TITLE: 0.??.? (2026-??-??)
+
+[TEXT]
+MID: 0046268453bf439b8a826ecc57abbe91
+STATEMENT: >>>
+This release contains the following enhancements:
+
+?\) The Markdown format now supports ``Version``, ``Date`` and ``Classification`` as document config fields.
+<<<
+
+[[/SECTION]]
+
+[[SECTION]]
 MID: 8a4b0e6f9d3c4c5b8e1a2f7d6c9b0a3e
 TITLE: 0.27.0 (2026-07-13)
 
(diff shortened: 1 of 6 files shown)


docs/strictdoc_04_release_notes.sdoc
```diff
@@ -49,6 +49,20 @@ releases:
 `StrictDoc Releases <https://github.com/strictdoc-project/strictdoc/releases>`_.
 <<<
 
+[[SECTION]]
+MID: a7ffb392d48244969318ec115e4695bd
+TITLE: 0.??.? (2026-??-??)
```
stanislaw, line 54: FYI: We usually call this `Unreleased` without a date initially. But I don't want this to block the merging of this PR.

docs/strictdoc_04_release_notes.sdoc
```diff
@@ -49,6 +49,20 @@ releases:
...
+
+[TEXT]
+MID: 0046268453bf439b8a826ecc57abbe91
+STATEMENT: >>>
+This release contains the following enhancements:
+
+?\) The Markdown format now supports ``Version``, ``Date`` and ``Classification`` as document config fields.
```
stanislaw, line 61: Here you could simply write `1\)` but again not a blocker.

strictdoc/backend/markdown/reader.py
```diff
@@ -340,6 +340,12 @@ def _parse_document_root(
                     document.config.requirement_prefix = field_.value
                 if field_.name == "UID":
                     document.config.uid = field_.value
+                if field_.name == "Version":
+                    document.config.version = field_.value
+                if field_.name == "Date":
+                    document.config.date = field_.value
+                if field_.name == "Classification":
```
stanislaw, line 347: FYI: I was thinking of deprecating the Classification field by the end of this year because we created it long time ago when there was no custom metadata supported. Now `Classification` seems to be custom enough and I think we could remove it to clean up the core model a little.

stanislaw (approved):

PR #3057, opened 2026-07-23
Update XlsxWriter version constraint to 3.*

cperkulator:
UV caught a version incompatibility on a project.  Hoping this update is ok


stanislaw:
Thanks! I only had to edit your commit so that it conforms to the Conventional Commits convention required by our linting checks.

PR #3065, opened 2026-07-26
fix(export/html): defer initial MathJax/Mermaid rendering to browser idle time

stanislaw:
WHY: Ensures that the rendering of MathJax/Mermaid does not block the first paint. This change will be useful for rendering large documents when we implement lazy loading in the editable UI (coming in the next PR #3049).


stanislaw:
cc @martinezlc99

PR #3071, opened 2026-07-27
feat(server): lazily load large documents in chunks in the web GUI 

mettta:
This PR includes the chunked-document loading implementation introduced
in [lazily load large documents in chunks in the web GUI](https://github.com/strictdoc-project/strictdoc/commit/8543abebff63f0630174ab207a06560e3f062101)  by @[martinezlc99](https://github.com/strictdoc-project/strictdoc/commits?author=martinezlc99):

> Documents with more than chunked_documents_threshold nodes (default
> 200, 0 disables) render the first chunk inline and load the remaining
> content as lazy turbo-frames while scrolling, served by a cursor-based
> fragment endpoint. TOC clicks, URL deep-links, and back/forward
> navigation eagerly load the chunk containing their target anchor.
> Non-chunked documents and the static HTML export are unaffected.

----

Building on this implementation, this PR adds:
- stable viewport positioning while document geometry changes
- correct TOC navigation and highlighting across loaded and unloaded chunks
- reliable viewport behavior during document editing.

### Problem

Lazy loading changes document geometry over time. A placeholder can have a
different height from the rendered chunk that replaces it. Editing can also
replace the complete content DOM and turn previously loaded content back into
placeholders.

Without extra handling, content already visible to the user can move when a
chunk above the viewport loads or when an editing operation returns updated
document content.

### TOC navigation and highlighting

TOC navigation now works whether the target is already rendered or belongs to
an unloaded chunk.

When needed, the target chunk is loaded before scrolling to it. Navigation
still uses the browser's normal URL fragment behavior, including `hashchange`,
CSS `:target`, and browser history.

TOC scrolling is immediate rather than smooth. This prevents the scroll from
ending at the wrong position if placeholders between the current position and
the target change height during the scroll.

Current-section highlighting now reacts to anchors added after the initial page
load. It uses the actual `.main` scroll container for its viewport geometry and
continues working when editing replaces an anchor with a new DOM element that
has the same ID.

TOC highlighting was also optimized for large documents. It uses ordered anchor
caches and binary search instead of reading the position of every loaded anchor
on each update. When a lazy chunk is inserted, only its new anchors are
registered. Complex DOM updates and same-ID replacements still use a full
refresh as a safe fallback.

### Stable content viewport

A new viewport controller keeps the content currently viewed by the user in the
same visible position while document geometry changes.

Before a relevant DOM update, the controller records a visible document element
and its position inside the content viewport. After the update, it finds the
same element in the new DOM and corrects the scroll position by the distance
that the element moved.

This preserves the actual document content being viewed instead of preserving
a raw `scrollTop` value.

The controller handles:

- lazy placeholders being replaced by rendered chunks;
- removal of the placeholder's estimated height;
- complete content replacement after server operations;
- additional chunk loads caused by the replacement;
- later height changes inside loaded chunks;
- create, delete, move, and grammar-edit operations;
- both chunked and non-chunked documents.

Different actions can request different viewport positions at the same time.
TOC and URL fragment navigation have priority because they request a specific
destination. Operations such as creating a node can also provide a specific
target. Otherwise, the controller keeps the content already visible to the
user in place.

New user input cancels older pending restoration work. A late chunk response
must not return the user to a position recorded before they scrolled somewhere
else.

The controller also treats active scrolling differently from an idle viewport.
While the user is continuously scrolling, passive chunk loading is left to the
browser's native scroll anchoring. Applying an exact correction from the
previous scroll frame could otherwise cause a small jump in the opposite
direction. Exact viewport preservation is used again after scrolling stops.

### Document editing

Creating a node can replace the full content DOM and place the new node in a
chunk far away from the form where the action started. The new node may also
have no TOC entry.

Lazy placeholders now contain the MIDs of all nodes in their chunks. This lets
the viewport controller find and load the exact chunk containing the created
node without depending on TOC data.

After document operations:

- a created node appears at the position of the creation form and stays there
  while surrounding chunks finish loading;
- deleting a visible node keeps the boundary from which the node was removed;
- moving nodes keeps the current viewport stable;
- editing the document grammar keeps the current viewport stable;
- TOC and URL fragment navigation still finish at the requested target.

### Placeholder estimates and preloading

Unloaded chunks use placeholders with estimated heights. Chunks also begin
loading shortly before they enter the viewport.

The estimated height gives unloaded content approximate space in the document,
and preloading reduces the chance that the user reaches an empty chunk. These
mechanisms improve loading behavior, but the viewport does not depend on the
height estimate being accurate. If the rendered chunk is much taller or shorter
than its placeholder, the viewport controller compensates for the difference.

### Test coverage

The test coverage includes:

- viewport stability after a chunk above the viewport loads;
- stability during the browser paint interval, not only after loading settles;
- continuous wheel scrolling in both directions;
- near-simultaneous loading of multiple chunks;
- delayed height changes inside loaded chunks;
- creating nodes in distant unloaded chunks without TOC entries;
- delete-boundary behavior;
- create, delete, move, and grammar editing in chunked and non-chunked
  documents;
- TOC navigation and highlighting for loaded and unloaded targets;
- TOC performance on large documents.


developer/tasks/20260728_pr3071_testing_lazy_loading/task_1_toc_navigation_and_highlighting.md
```diff
@@ -0,0 +1,145 @@
+# TOC navigation and highlighting in chunked documents
+
+## WHAT
+
+TOC current-section highlighting, deep-link navigation (TOC click,
```
stanislaw, line 5: Let's have a more smooth introduction to the WHAT of this work. Which behavior are implementing here?

developer/tasks/20260728_pr3071_testing_lazy_loading/task_1_toc_navigation_and_highlighting.md
```diff
@@ -0,0 +1,145 @@
...
+Server-side lazy loading of large documents in chunks (PR #3049)
+inserts content into the DOM well after initial page load, and drives
+deep-link navigation programmatically. Both TOC highlighting and TOC
+navigation predate chunking and were built assuming all content - and
+the TOC itself - are fully present and synchronized from the first
+`load` event. Chunking breaks that assumption in three independent
+ways:
+
+1. `toc_highlighting.js` only re-scans content anchors when the TOC
+   frame itself mutates. Before chunking, almost every content change
```
stanislaw, line 62: e.g., when a content node is edited or deleted,

developer/tasks/20260728_pr3071_testing_lazy_loading/task_1_toc_navigation_and_highlighting.md
```diff
@@ -0,0 +1,145 @@
...
+   was accompanied by a TOC-frame mutation (the server always
+   re-renders and streams the full TOC alongside any content change) -
+   Cancel was already a narrow, pre-existing exception (see
+   `../20260726_toc_highlight_improvement/task.md`). Lazily-loaded
+   chunks are the primary case where content changes without touching
```
stanislaw, line 67: This sentence is a little hard to read.

developer/tasks/20260728_pr3071_testing_lazy_loading/task_1_toc_navigation_and_highlighting.md
```diff
@@ -0,0 +1,145 @@
...
+   target's actual position mid-scroll - the animation then lands
+   short of it.
+3. CSS `:target` and the `hashchange` event are both tied to the
+   browser's actual fragment-navigation processing. Driving navigation
+   programmatically (needed to force-load a chunk before scrolling to
+   it) can silently produce neither, unless done carefully.
+
+## HOW
+
+**`toc_highlighting.js`** (see
```
stanislaw, line 83: At this point, it is not immediately clear whether this is talking about the new or the old behavior.

stanislaw: One way of improving this: we are moving from A (old) to B (new).

developer/tasks/20260728_pr3071_testing_lazy_loading/task_1_toc_navigation_and_highlighting.md
```diff
@@ -0,0 +1,145 @@
...
+`../20260726_toc_highlight_improvement/task.md` for the original,
+detailed writeup): react to newly-inserted content anchors via the
+shared `StrictDoc.onInsert` contract, in addition to the existing
+TOC-frame observer.
+
+The steady-state scroll path keeps an ordered cache of anchors that have TOC
+links. Section intervals are vertically ordered, so two binary searches locate
+the interval overlapping the viewport without reading every loaded anchor's
+geometry. Only TOC links whose `intersected` state changed receive an attribute
+write. A 600-anchor regression records DOM geometry reads rather than unstable
```
stanislaw, line 93: is this referring to a regression test?

developer/tasks/20260728_pr3071_testing_lazy_loading/task_1_toc_navigation_and_highlighting.md
```diff
@@ -0,0 +1,145 @@
...
+wall-clock time: the previous full scan performed 1199 reads for one update;
+the logarithmic implementation performs 19, with a browser-tolerant ceiling of
```
stanislaw, line 95: Test fixture (600 anchors): Before optimization: 1119 reads, after optimization: 19 with a test constraints that the number of reads should be less than 64.

strictdoc/export/html/_static/content_viewport_restoration.js
```diff
@@ -0,0 +1,778 @@
+(() => {
```
stanislaw, line 50: Would be nice to have the Bottom Line Up Front here.

strictdoc/export/html/_static/content_viewport_restoration.js
```diff
@@ -0,0 +1,778 @@
...
+  }
+
+  document.addEventListener("click", (event) => {
+    const confirmLink = event.target.closest?.(
+      "a[data-testid='confirm-action']"
+    );
+    if (!confirmLink) return;
+
+    const actionUrl = new URL(confirmLink.href, document.baseURI);
+    if (!actionUrl.pathname.endsWith("/delete_requirement")) return;
```
stanislaw, line 1051: Generalize the selectors

strictdoc/export/html/_static/content_viewport_restoration.js
```diff
@@ -0,0 +1,778 @@
...
+    restoreViewportAnchor(
+      pendingSnapshot.snapshot,
+      pendingSnapshot.generation
+    );
+    observeChunkGeometry(frame, pendingSnapshot);
+  });
+
+  document.addEventListener("turbo:fetch-request-error", (event) => {
+    const frame = event.target;
+    if (!frame?.id?.startsWith("document-chunk-")) return;
```
stanislaw: Generalize, add constant.

strictdoc/export/html/_static/content_viewport_restoration.js
```diff
@@ -0,0 +1,778 @@
+(() => {
+  const strictDoc = window.StrictDoc;
```
stanislaw, line 52: What exactly does this do to ensure a browser-like viewport stabilization?

strictdoc/export/html/_static/draggable_list.js
```diff
@@ -249,7 +249,12 @@ ${SEL_ITEM}[draggable="true"] .dragIndicator::before {
           'Accept': 'text/vnd.turbo-stream.html',
         },
       }).then((r) => r.text())
-      .then((html) => Turbo.renderStreamMessage(html));
+      .then((html) => {
+        // Move uses manual fetch + Turbo stream rendering. The shared
+        // before-stream-render handler captures immediately before the
+        // content frame is replaced.
+        window.StrictDoc.contentViewport.renderManualStreamMessage(html);
```
stanislaw: The `renderManualStreamMessage` seems to be calling `renderStreamMessage` only and nothing else. Is it intentional or we could simply remove this indirection?

strictdoc/export/html/templates/screens/document/document/_node_dispatch.jinja.html
```diff
@@ -0,0 +1,11 @@
+        {%- if node.is_document() %}
```
stanislaw: Let's remove the indentation in this file.

developer/tasks/20260728_pr3071_testing_lazy_loading/task_1_toc_navigation_and_highlighting.md
```diff
@@ -0,0 +1,145 @@
+# TOC navigation and highlighting in chunked documents
```
stanislaw, line 1: I suggest that we significantly reduce the size of the PR description to only minimal key points, mostly referring to the `developer/tasks` files. Also, we wanted to write a short main task that would summarize the key aspects of what was added in the first commit.

strictdoc/export/html/templates/screens/document/document/_node_dispatch.jinja.html
```diff
@@ -0,0 +1,11 @@
...
+          {% include "components/section/index_extends_node.jinja" %}
+        {%- elif node.is_requirement() %}
+          {# {%- if node.is_text_node() %}
```
stanislaw: Should we remove the commented out lines?

stanislaw:
@martinezlc99, your work has been merged, along with @mettta's changes for TOC and viewport handling.

We are leaving the documentation work for a separate PR. We want to document this feature at the L2 and L3 specification levels and remove the implementation-level documentation that is currently located under `developer/tasks` for this PR.

Issue #3050, opened 2026-07-20
Feature: Test reports: map JUnit testcase `<properties>` into TEST_RESULT nodes (enables attaching test evidence such as plots)


johanenglund:
## Problem

We ingest GoogleTest JUnit XML into StrictDoc as an extra input path (`strictdoc export . <test_report_dir>`), and the generated TEST_RESULT nodes give us test-to-requirement traceability in the HTML report. This works well.

What we cannot do is carry any test **evidence** along with a result. Several of our tests produce artifacts that a reviewer wants to see next to the pass/fail status, today those artifacts have no route into the report.

JUnit XML has a natural, framework-neutral slot for this: testcase `<properties>`. GoogleTest writes them via `RecordProperty()`, pytest via `record_property`, and most other frameworks have an equivalent. But `junit_xml_reader.py` currently extracts only name/classname/file/line/time/status per testcase — `<properties>` (and `<system-out>`) are discarded.

## Proposal

1. **Generic mapping:** for each `<property name="..." value="..."/>` inside a `<testcase>`, add a corresponding field to the generated TEST_RESULT node (and to the auto-generated grammar of the test-report document), so arbitrary tool-side metadata becomes visible in the report.

2. **An attachment/image convention on top:** treat a reserved property name (e.g. `strictdoc:image` or `strictdoc:attachment`) as a file reference — path resolved relative to the JUnit XML file's location, asset copied into the export output, and rendered inline (image) or as a link in the TEST_RESULT node in the HTML report.

Example of what the producing side already looks like with plain GoogleTest:

```cpp
TEST(TestSuite, Test1)
{
    // ... perform tests, writes test1.png next to the XML ...
    RecordProperty("strictdoc:image", "test1.png");
}
```

```xml
<testcase name="Test1" ... >
  <properties>
    <property name="strictdoc:image" value="test1.png"/>
  </properties>
</testcase>
```

Item 1 alone would already be valuable (we could post-process or template around it); item 2 is the part that makes the report self-contained evidence.

## Environment

- strictdoc 0.26.0
- JUnit XML produced by GoogleTest (`--gtest_output=xml:`)

PR #3072, opened 2026-07-27
feat(test-reports): map JUnit testcase properties

LunaMeerkats:
## What changed

- Map testcase `<properties>` into generated `TEST_RESULT` fields.
- Normalize property names to valid uppercase StrictDoc field identifiers while
  preserving the original names as human-readable field titles.
- Reject invalid, reserved, normalization-colliding, and duplicate property
  names with actionable errors.
- Add focused unit and HTML integration coverage.
- Add an Unreleased note.

## Why

StrictDoc previously discarded testcase properties during JUnit XML ingestion,
so test-runner metadata could not appear in generated reports. This implements
the generic metadata portion of #3050 without defining attachment copying or
special rendering semantics.

Duplicate property names remain valid across different testcases. They are
rejected within one testcase because StrictDoc fields are single-valued and a
silent last-value-wins mapping would lose metadata emitted by runners such as
pytest.

## User impact

JUnit metadata such as requirement identifiers and `strictdoc:image` now appears
as ordinary string fields in generated test-result reports. Existing reports
without testcase properties are unchanged.

## Validation

- `invoke test-unit`: 655 passed, 1 skipped.
- `invoke lint-mypy`: passed.
- `invoke lint-ruff`: passed.
- `invoke test-integration --focus 06_testcase_properties
  --no-parallelization` under WSL: 1 selected test passed.
- `invoke docs`: the changed release notes parsed successfully; the export then
  stopped on the existing Windows path-length limitation.

### DIFF
--- a/docs/strictdoc_04_release_notes.sdoc
+++ b/docs/strictdoc_04_release_notes.sdoc
@@ -50,6 +50,21 @@ releases:
 <<<
 
 [[SECTION]]
+MID: 295e21753ef346d88c2ccb0ceecd0657
+TITLE: Unreleased
+
+[TEXT]
+MID: c9ceba5f4cc74d04b7f389962b1f127a
+STATEMENT: >>>
+This release contains the following enhancements:
+
+1\) The JUnit XML test report reader now maps testcase ``<properties>`` to
+fields on generated ``TEST_RESULT`` nodes.
+<<<
+
+[[/SECTION]]
+
+[[SECTION]]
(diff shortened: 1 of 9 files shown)


stanislaw (approved):
@LunaMeerkats I think everything looks good. I see the Draft marker - Do you want to make any further changes?

stanislaw:
Actually I checked this again, and I think it is mergeable as-is. Thanks for contributing :pray:

Issue #2232, opened 2025-05-08
Feature: JUnit XML -> SDoc verification report generator: mark failing test's corresponding requirements accordingly


MartyLake:
## Description

I tried the JUnit XML -> SDoc generator with ctest xml, and it works great!

> children: HelloTest.BasicAssertions_Float HelloTest.BasicAssertions_Float (IsSatisfiedBy)

## Problem

I find it hard to see the consequences of a failing test. My toy project is trivial, but I imagine a longer chain of requirements, and I wish I could transitively see the upstream consequences of a failing test from the requirements POV.

## Solution

The corresponding test's requirement, and its parents, could have an indicator that a downstream test is failing. This indicator could be used to understand which areas of the software needs attention during manual testing, for example.

LunaMeerkats:
One detail seems worth settling before I turn this into a regression. On current main, a requirement can reach a result either directly (`is satisfied by`) or through a generated test case (`is verified by` → `has result`). `TEST_RESULT` is also used by Robot (`PASS`/`FAIL`/`SKIP`) and gcov (`Covered`/`Non-covered`), while JUnit uses `PASSED`/`FAILED`/`SKIPPED` and currently rejects `<error>` results.

My inclination is to preserve that provenance: the verified requirement would have a direct failed result, while its parent would be marked only as affected by a downstream failure, without changing either requirement’s own `STATUS`. Would you prefer a JUnit-only first step, or should this wait until StrictDoc has a normalized result kind/outcome?

PR #3088, opened 2026-08-06
feat(traceability): preserve verification result provenance

LunaMeerkats:
This adds a narrow traceability query that preserves where verification results came from. It is a foundation for #2232, without choosing result aggregation or presentation semantics yet.

The new `VerificationResultProvenance` record keeps the requirement actually verified, an optional generated test case, and the raw result node. `TraceabilityIndex.get_verification_result_provenance()` follows the two graph shapes StrictDoc already creates:

- `IS_SATISFIED_BY` for a direct requirement-to-result link;
- `IS_VERIFIED_BY` followed by `HAS_RESULT` for a generated test-case link.

The query also traverses ordinary requirement hierarchy in deterministic order. This lets a consumer distinguish a result attached to the selected requirement from one attached to a descendant. It deliberately does not normalize producer-specific statuses, mutate a requirement's own `STATUS`, aggregate pass/fail state, or add HTML indicators.

The integration fixture builds the real SDoc, C++, and CTest/JUnit graph. It covers direct and mediated links, a two-level requirement hierarchy, exact ordering, duplicate prevention, raw failed-result preservation, and unchanged requirement statuses.

Validation:

- `invoke test-integration --focus verification_result_provenance --no-parallelization`
- `invoke test-unit --path tests/unit/strictdoc/core/test_traceability_index.py`
- targeted Ruff format and lint checks
- `invoke lint-mypy`
- `invoke lint-commit`

### DIFF
--- a/docs/strictdoc_04_release_notes.sdoc
+++ b/docs/strictdoc_04_release_notes.sdoc
@@ -63,6 +63,10 @@ fields on generated ``TEST_RESULT`` nodes.
 
 4\) Fixed: the TABLE VIEW now renders the Relation cell as non-editable for node
 types whose grammar does not declare relations.
+
+5\) The traceability index now preserves whether each result verifies the
+selected requirement or one of its descendants, and whether a generated test
+case mediates the link.
 <<<
 
 [[/SECTION]]
--- a/strictdoc/core/traceability_index.py
+++ b/strictdoc/core/traceability_index.py
@@ -4,6 +4,7 @@
 
 import datetime
 from copy import copy, deepcopy
+from dataclasses import dataclass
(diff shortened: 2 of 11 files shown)


LunaMeerkats:
Main has moved through 0.28.1 since this opened, and GitHub now reports conflicts in `traceability_index.py` and the release notes. I checked current main: no equivalent provenance query has landed, and the graph APIs used by this change are still present. Resolving the conflicts requires a one-commit rebase that combines the import changes and moves the note into the current Unreleased section.

I do not rewrite published branch history. I can open a replacement PR from current main; a maintainer would then need to close this one. Please confirm if you want me to take that route.

PR #3095, opened 2026-08-10
feat(export/html): stabilize scroll position for lazy chunks in static export

mettta:
Enabled scroll stabilization for chunked static documents. To do this, we manually dispatch a `turbo:before-fetch-response`  event before splicing in the chunk content - `content_viewport_restoration.js` needs it to snapshot the viewport position before geometry changes.


stanislaw (approved):

stanislaw:
cc @martinezlc99 this is now merged as well.

Issue #2061, opened 2025-01-17
Consider migrating to uv Python package/project manager


stanislaw:
Our friendly open source project is reporting very nice things about the uv project: https://github.com/DSD-DBS/py-capellambse/pull/494#pullrequestreview-2527580284.

Consider switching to uv.

nogjam:
You might consider the alternative viewpoint as well. These people say it better than I can: https://mastodon.social/@glyph/113093806840686501

In short, the tool itself seems to function very nicely. But sustainability-wise, it may have some bad marks against it.

I use [Poetry](https://python-poetry.org/) in most of my projects.

PR #3108, opened 2026-08-15
feat(nix): build and run StrictDoc from the flake

adamatom:
Before this change, the flake gave a Python interpreter and system libraries, but not the Python dependencies. The shell installed those with pip each time it started, into a virtual environment .venv-nix. The file .envrc made a second virtual environment, .venv. The result was a different environment for each way to enter the directory.

Now pyproject.nix reads pyproject.toml and finds each dependency in nixpkgs. The shell gives the environment that the project declares. The pip step and both virtual environments are no longer necessary. The file .envrc contains only `use flake`, thus direnv and `nix develop` give the same environment.

The flake also builds StrictDoc. Thus you can run it without a copy of the source:

    nix run github:strictdoc-project/strictdoc -- --help

The flake contains four package definitions. nixpkgs does not have tree-sitter-c and tree-sitter-cpp. It has reqif and python-datauri, but with a version that pyproject.toml does not permit.

The interpreter is Python 3.13, and not 3.12. nixpkgs supplies binaries for its default interpreter and for the one below it. With 3.12, many of the python packages would need to built from source.

The shell installs github_changelog_generator directly in place of ruby, since it was already available in nixpkgs.

---

You can test this off of my fork with:

    nix run github:adamatom/strictdoc/complete_flake -- --help

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -4,90 +4,191 @@
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
     flake-utils.url = "github:numtide/flake-utils";
+
+    pyproject-nix = {
+      url = "github:pyproject-nix/pyproject.nix";
+      inputs.nixpkgs.follows = "nixpkgs";
+    };
   };
 
-  outputs = { self, nixpkgs, flake-utils }:
-    flake-utils.lib.eachDefaultSystem (system:
+  outputs =
+    {
+      nixpkgs,
+      flake-utils,
+      pyproject-nix,
+      ...
+    }:
+    flake-utils.lib.eachDefaultSystem (
+      system:
       let
         pkgs = import nixpkgs { inherit system; };
 
-        # Built by nixpkgs itself -- numpy/pandas's compiled extensions
-        # are already correctly linked against Nix's own libstdc++ (no
-        # patchelf needed for these specifically). Symlinked directly
-        # into .venv-nix's site-packages in shellHook below, instead of
-        # pip-installing this whole stack from PyPI. six/python-dateutil/
-        # pytz/tzdata are pandas's own runtime deps, pulled in the same
-        # way so pip never needs to touch any of this.
-        nixNumpyStack = with pkgs.python312Packages; [
-          numpy
-          pandas
-          six
-          python-dateutil
-          pytz
-          tzdata
-        ];
+        # This pulls in all of the python dependencies defined by
+        # pyproject.toml, but none from tox.ini. This means we have enough
+        # to run strictdoc, but not the developer tooling.
+        project = pyproject-nix.lib.project.loadPyproject { projectRoot = ./.; };
+
+        # Use 3.13 for the environment because the nixpkgs cache contains
+        # cached builds for some of the larger python dependencies like
+        # scipy. When using 3.12, it requires building scipy from source.
+        python = pkgs.python313.override {
+          self = python;
+          packageOverrides = pkgs.lib.composeExtensions treeSitterOverrides versionOverrides;
+        };
+
+        # pyproject.toml pins versions that are not in nixpkgs, so we override
+        # the versions with the following. If strictdoc adopts uv, then the
+        # uv2nix project (which is a subproject of pyproject-nix), can use
+        # the uv.lock to automatically resolve these versions.
+        versionOverrides = final: prev: {
+          reqif = prev.reqif.overrideAttrs (_old: rec {
+            version = "0.1.0";
+            # From the repository rather than the sdist, because nixpkgs' own
+            # derivation patches a file under tests/, which the sdist omits.
+            src = pkgs.fetchFromGitHub {
+              owner = "strictdoc-project";
+              repo = "reqif";
+              tag = version;
+              hash = "sha256-aMjq2x9/aC7HRDL2T2v/yvz+TP+AAKSY3e/TmboKq9Q=";
+            };
+          });
+
+          python-datauri = prev.python-datauri.overrideAttrs (_old: rec {
+            version = "2.2.0";
+            src = final.fetchPypi {
+              inherit version;
+              pname = "python_datauri";
+              hash = "sha256-j6WCIpdf7lQc455rC51WU878wkga0mKe/AN41IsT9Ds=";
+            };
+
+            # tests/assets holds fixture files, one of which is EBCDIC-encoded
+            # and named test_*, so pytest tries to collect it as a module and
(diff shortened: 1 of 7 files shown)


.envrc
```diff
@@ -1,9 +1 @@
-#
```
stanislaw, line 1: I was going to maintain the current setup and Nix setup in parallel for some time. I want to first explain Nix to the other contributors (especially @haxtibal and also @puzzlewolf who created a Nix package before) during the Office Hours and via the mailing list. Side note: The use of direnv (I learned about it from you some time ago) has not been documented yet.

Let's keep `.envrc` and `bootstrap_nix.sh` as they are for some time. I see this as an acceptable approach for some time since this new flake already enables StrictDoc to be installed from another Nix environment. I myself want to start using the new Nix flake to better understand where I want to migrate the rest of the infrastructure (likely: remove Tox, introduce uv, make Nix default).

Also, we should document this workflow in the User Guide around https://strictdoc.readthedocs.io/en/latest/latest/docs/strictdoc_01_user_guide.html#3.6-Installing-StrictDoc-with-Nix and also in the Developer Guide. The documentation part I am happy to do by myself once we merge this one or feel free to do it.

flake.nix
```diff
@@ -4,90 +4,153 @@
...
-        # GLIBC_* errors. Separate venv directories avoid that entirely.
+        # This shell provides the Python environment the project declares,
+        # plus the system-level tools that are not Python packages at all:
+        # libtidy (pytidylib, integration tests), a browser
+        # (seleniumbase/playwright end2end and screencast tests), and
+        # github_changelog_generator.
         devShells.default = pkgs.mkShell {
           packages = [
-            pkgs.python312
+            pythonEnv
```
stanislaw, line 153: Now that I better understand the separation and difference between `devShells` and `packages.default = strictdoc` 😄, could we add `invoke` and `tox` here so that all Invoke tasks can be run from this flake when we enter it with `bootstrap_nix.sh`?

adamatom: I have added the tox integration. I was able to keep the pyproject.toml dependencies installed into the nix environment and have tox use it, and then only add the development dependencies on-top and outside of the nix environment.

I recommend switching to `uv`, because you can then use `uv2nix`, which is a subproject of `pyproject.nix` used here. The `uv.lock` is then enough to make `nix` happy, each entry contains a hash from the PyPI, and you can get the best of both `uv` and `nix`. You get to use `uv` to manage what goes into your `pyproject.toml`, and fully defined packages though `uv.lock`, and then `nix` installs everything fully defined. The people who do not want to use `nix` can fall back to your `uv.lock` and get the same python environment at least.

.envrc
adamatom: Ah good points. I saw .envrc and assumed it was just an awkward way of getting ready to use the flake. I reverted `.envrc`.  `bootstrap_nix.sh` is just a wrapper around `nix develop` that a nix, but seems harmless.

I have some notes about how to use this flake now, one set is for nix users of strictdoc, and the other is for developers, I haven't written them into the user guide. This needs to be woven into the nixpkgs package description, since this is now an alternative for nix people who use flakes:

For users:

```
> ## Run StrictDoc without installing it
> 
> Run StrictDoc directly from its flake:
> 
> ```bash
> nix run github:strictdoc-project/strictdoc
> ```
> 
> Pass arguments to StrictDoc after `--`:
> 
> ```bash
> nix run github:strictdoc-project/strictdoc -- export docs/
> nix run github:strictdoc-project/strictdoc -- --help
> ```
> 
> Nix downloads StrictDoc and its dependencies and runs the command without
> adding StrictDoc to your PATH.
> 
> ## Start a shell with StrictDoc available
> 
> ```bash
> nix shell github:strictdoc-project/strictdoc
> strictdoc export docs/
> ```
> 
> The `strictdoc` command remains available until you leave the shell.
> 
> ## Select a release
> 
> Append a release tag to the flake URL:
> 
> ```bash
> nix run github:strictdoc-project/strictdoc/<version>
> ```
> 
> For example:
> 
> ```bash
> nix run github:strictdoc-project/strictdoc/0.XX.X
> ```
> 
> ## Add StrictDoc to your own flake
> 
> Add StrictDoc as a flake input:
> 
> ```nix
> {
>   inputs.strictdoc.url = "github:strictdoc-project/strictdoc";
> 
>   outputs = { nixpkgs, strictdoc, ... }: {
>     # ...
>     environment.systemPackages = [
>       strictdoc.packages.x86_64-linux.default
>     ];
>   };
> }
> ```
> 
> If your flake already has a `nixpkgs` input, you can make StrictDoc use the
> same `nixpkgs` revision:
> 
> ```nix
> inputs.strictdoc = {
>   url = "github:strictdoc-project/strictdoc";
>   inputs.nixpkgs.follows = "nixpkgs";
> };
> ```
> 
> This avoids StrictDoc using a separate `nixpkgs` input and keeps the
> dependency graph aligned with your own flake.
```

For developers:

```

> ## Enter the development shell
> 
> Clone the repository and enter the Nix development shell:
> 
> ```bash
> git clone https://github.com/strictdoc-project/strictdoc.git
> cd strictdoc
> nix develop
> ```
> 
> Alternatively, run:
> 
> ```bash
> ./bootstrap_nix.sh
> ```
> 
> The script enters the same development shell and prints the Nix command it
> runs.
> 
> ## Run the developer tasks
> 
> List the available Invoke tasks:
> 
> ```bash
> invoke --list
> ```
> 
> Run the main checks:
> 
> ```bash
> invoke test-unit
> invoke lint
> invoke test-integration
> invoke test-end2end --headless
> ```
> 
> The first Tox-based task can take a few minutes and requires a network
> connection. Tox creates its virtual environment in `build/tox/` and installs
> the development dependencies defined by `tox.ini` from PyPI. Subsequent runs
> reuse the environment and start more quickly.
> 
> StrictDoc's runtime dependencies are provided by Nix rather than installed
> from PyPI. The development shell adds the Nix environment's `site-packages` to
> the `PYTHONPATH` used by the Tox environments.
> 
> As a result, `developer/pip_install_strictdoc_deps.py` finds the runtime
> dependencies already available and does not install them again.

```

stanislaw: This is great. I will make sure to integrate these instructions as-is during my next pass on StrictDoc.

flake.nix
stanislaw: I have added your comment directly to the uv backlog ticket: https://github.com/strictdoc-project/strictdoc/issues/2061.

stanislaw (approved):

PR #3131, opened 2026-08-20
Feature: Move nodes within and across documents

mettta:
Adds a project-wide tree for moving complete node subtrees while preserving their structure, fields, and relations.

_Moving is currently disabled for subtrees containing images or autogenerated nodes, and cross-document moves are blocked when the destination grammar cannot represent the subtree unchanged._


stanislaw:
@thseiler FYI this work is entering the review. Our next stop after merging this will be to integrate your Image PR.

After several discussions, both related and unrelated to your PR, we realized that we do want to proceed with `@assets` macro because it gets pretty difficult to handle images otherwise.

stanislaw (approved):

Issue #3129, opened 2026-08-19
Feature: Enable Composable Prefixes for UIDs


pmo73:
## Description

I would like to request support for defining prefixes for individual types in the StrictDoc grammar while also allowing an additional document-level prefix to be specified.

The resulting UID should be composed of the document prefix, the type-specific grammar prefix, and the UID number.

For example, a project-wide grammar could define the following prefixes:

    Requirements → REQ
    Definitions → DEF
    Assumptions → ASP

A document could then define an additional prefix such as SDD, resulting in UIDs like:

    SDD-REQ-1
    SDD-DEF-1
    SDD-ASP-1

Another document could use the prefix SDD-LLR, resulting in UIDs such as:

    SDD-LLR-REQ-1
    SDD-LLR-DEF-1

This would allow the same grammar to be reused across an entire project while still providing document-specific prefixes.

## Problem

Currently, the prefix defined in the grammar overrides the prefix specified at document level. If no prefix is specified in the grammar, the document-level prefix is used instead.

This makes it difficult to use a single, project-wide grammar while also distinguishing UIDs based on the document they belong to.

For larger projects, it would be useful to define the prefixes for all types centrally in one grammar and then have each document add its own prefix. Currently, achieving this would require creating or modifying a grammar specifically for each document, even though the type definitions and their prefixes are identical across the project.

## Solution

I suggest extending the prefix handling so that the grammar-level type prefix and the document-level prefix are combined instead of one overriding the other.

The resulting UID would follow this pattern:

<document-prefix>-<grammar-type-prefix>-<uid-number>

For example:

Grammar:
  Requirement → REQ
  Definition  → DEF
  Assumption  → ASP

Document:
  Prefix → SDD

would result in:

SDD-REQ-1
SDD-DEF-1
SDD-ASP-1

A different document could specify:

Document:
  Prefix → SDD-LLR

resulting in:

SDD-LLR-REQ-1
SDD-LLR-DEF-1
SDD-LLR-ASP-1

The expected behavior could be:

1. The prefix of a type is defined in the grammar.
1. A document can optionally define an additional prefix.
1. If both prefixes are present, they are concatenated.
1. If only the grammar prefix is present, it is used on its own.
1. If only the document prefix is present, it is used on its own.
1. If neither is present, the existing behavior should be preserved.

## Additional Information

The main use case is having a single grammar for an entire project, with centrally defined prefixes for the different requirement/node types. The document-level prefix can then identify the specific document or document context.

[...]

PR #3133, opened 2026-08-20
feat(project_config): export a user-provided custom node prefix

stanislaw:


stanislaw:
@pmo73 @SLeeUVAD

I am proposing that we solve both use cases by adding user hooks to the ProjectConfig:

1) The custom prefixes like is requested here: https://github.com/strictdoc-project/strictdoc/issues/3129.

2) The `REQ-xxxx` placeholder, i.e., doing `REQ-0001` instead of `REQ-1`.

I will polish the PR docs some more. Let me know if you think this solves your use cases indeed.

(@SLeeUVAD, I just noticed your fork randomly and realized I was doing something very related yesterday).

SLeeUVAD:
The custom prefix concept would be quite useful. 

To provide more context, I wanted to implement the four-digit UID to make sorting smoother (e.g. my exporter would sort them 1, 10, 11 .... 19, 2, and so on).

While I only want 4 digits for my use case, allowing the user to modify the number of digits in the config file would be helpful as well.

SLeeUVAD:
Actually, now that you brought it up - having the requirement uids that follow custom prefix be appended with a '-' would be great for visibility. (e.g. REQ-TEST-0001 instead of REQ-TEST0001).

Would you know a quick way to achieve this? I am yet to understand the code structure fully, and it seems like there's a bit more to it than simple string concatenation.

stanislaw:
> Actually, now that you brought it up - having the requirement uids that follow custom prefix be appended with a '-' would be great for visibility. (e.g. REQ-TEST-0001 instead of REQ-TEST0001).
> 
> Would you know a quick way to achieve this? I am yet to understand the code structure fully, and it seems like there's a bit more to it than simple string concatenation.

I have added a test fixture to this PR that uses these two new hooks:

1) One hook helps to derive the prefix from document level prefix + grammar element type.

2) Another hook adds the `xxxx` padding.

Look at this test:

`tests/integration/features/commands/manage/auto-uid/requirement_uids/12_custom_prefix_hyphen_and_zero_padding/strictdoc_config.py`

stanislaw:
I will release this in a couple of days. In the meantime, you can try the new hooks by installing from the main branch directly:

```
```

SLeeUVAD:
Amazing, thank you!

SLeeUVAD:
One small thing, I suggest adding the same feature to cloning as well. It currently seems to be broken after the merge.

stanislaw:
@SLeeUVAD 

> One small thing, I suggest adding the same feature to cloning as well. It currently seems to be broken after the merge.

It is a good catch. I missed to add it to the clone function.

It should be fixed in this bugfix release, please give it a try:

https://github.com/strictdoc-project/strictdoc/releases/tag/0.28.3

PR #3141, opened 2026-08-22
feat(markdown): raise a validation error for sections with flat requirements

stanislaw:
As a workaround to avoid ambiguity, require a user add `TYPE: Section`.


stanislaw:
cc @haxtibal

PR #3155, opened 2026-08-26
feat(server): add support for drag-drop image upload (groundwork)

stanislaw:


LunaMeerkats:
Most PR #3155 test jobs stop before running their suites. The Linux HTML2PDF and end-to-end logs show the same startup error: `manage_assets_command.py` imports `WildcardEnhancedImage` from `strictdoc.export.rst`, but current `main` moved it to `strictdoc.backend.rst`.

Changing that import should let those jobs reach their tests. CodeQL is green, but the functional suites have not run far enough to validate this branch.

stanislaw:
Hey @LunaMeerkats, thanks for the heads up! The tests are failing expectedly because @mettta and I are integrating @thseiler's work with this separate PR. We have rebased the original work and now need to adapt it to the latest code structure.

Once this PR is integrated, we will do more work on this feature as well as proceed with integrating the other PRs.

docs/strictdoc_01_user_guide.sdoc
```diff
@@ -2757,10 +2757,33 @@ This is the example of how images are included using the reST syntax:
        :class: image
     <<<
 
+Sphinx may need different versions of the same image for different output
```
stanislaw: Here we should use the BLUP/WWH approach:

For users who want to export SDoc files to RST/Sphinx, StrictDoc extends an RST ``image`` directive to support wildcard-based paths. This extensions allows a user to store the same image in multiple formats. A typical use case: When Sphinx exports to HTML and PDF, it can use an SVG version of the image for HTML export and PNG version for PDF export.

stanislaw: Since this is a very custom extension inspired by Sphinx, I would even create a sub-section to contain it as child section here.

strictdoc/core/image_formats.py
```diff
@@ -0,0 +1,21 @@
...
+    ".svg": "image/svg+xml",
+    ".png": "image/png",
+    ".gif": "image/gif",
+    ".jpg": "image/jpeg",
+    ".jpeg": "image/jpeg",
+    ".webp": "image/webp",
+    ".avif": "image/avif",
+}
+
+SUPPORTED_IMAGE_FORMAT_NAMES = "SVG, PNG, GIF, JPG, JPEG, WebP, AVIF"
```
stanislaw, line 13: Let's derive this constant from `SUPPORTED_IMAGE_FORMATS`  for DRY.

strictdoc/core/image_formats.py
```diff
@@ -0,0 +1,21 @@
+from pathlib import Path
+
+SUPPORTED_IMAGE_FORMATS: dict[str, str] = {
```
stanislaw, line 3: Let's encapsulate all this content in an Enum class `ImageFormat`.

stanislaw: Also describe the use case in a docstring comment.

strictdoc/export/html/_static/editable_field.js
```diff
@@ -1,25 +1,381 @@
...
+  const SUPPORTED_IMAGE_FORMATS = new Map([
+    ['.svg', 'image/svg+xml'],
+    ['.png', 'image/png'],
+    ['.gif', 'image/gif'],
+    ['.jpg', 'image/jpeg'],
+    ['.jpeg', 'image/jpeg'],
+    ['.webp', 'image/webp'],
+    ['.avif', 'image/avif'],
+  ]);
+  const SUPPORTED_IMAGE_FORMAT_NAMES = 'SVG, PNG, GIF, JPG, JPEG, WebP, AVIF';
```
stanislaw, line 18: Here a comment that the source of truth is in `strictdoc/core/image_formats.py`.

strictdoc/server/routers/main_router.py
```diff
@@ -3470,6 +3706,7 @@ def document__add_comment(
                 existing_requirement_uid=None,
                 grammar=grammar,
                 relation_types=[],
+                document_markup=document.config.markup or SDocMarkup.RST,
```
stanislaw, line 3709: Instead of hardcoding this `or` decision here, we should use `document.config.get_markup()`. The same for the other use in this file.

PR #3156, opened 2026-08-27
feat(plantuml): add PlantUML diagram rendering support

### LINKED ISSUE #2791: Feature: Rendering Plant UML
### Discussed in https://github.com/strictdoc-project/strictdoc/discussions/2697

<div type='discussions-op-text'>

<sup>Originally posted by **PommeDeTerre0** February 25, 2026</sup>
Hey everyone,

In my company we're big fans of PlantUML diagrams to describe requirements, currently we render these diagrams separately and then add them to the requirements as images. 

I see that there is a backlog item for UML support, and I would like to have a go at it since I've implemented PlantUML rendering in some other projects in previous jobs. 

I'm not very familiar with the strictdoc code-base though, and some help to point me in the direction of where such implementation would be appropriate would allow me to get started much faster ! </div>

bachp:
Mirrors the existing Mermaid fence-rendering pattern (client-side JS injection replacing pre.plantuml/```plantuml blocks) for consistency with how diagramming tools are already integrated.

No default plantuml_server_url is configured: unlike Mermaid, which renders entirely in the browser, PlantUML rendering requires sending diagram source to a server, so the feature stays off until a project explicitly opts in and chooses a server it trusts with that content. Making it opt in with explicit server configuration prevents accidental information leakage.

Closes: https://github.com/strictdoc-project/strictdoc/issues/2791

### DIFF
--- a/docs/strictdoc_01_user_guide.sdoc
+++ b/docs/strictdoc_01_user_guide.sdoc
@@ -3961,6 +3961,69 @@ diagrams.
 [[/SECTION]]
 
 [[SECTION]]
+MID: 241e0e27e1c74262a97a9aaaefda6064
+TITLE: PlantUML diagramming tool
+
+[TEXT]
+MID: 6bd707bcbc2b4e77b65a8d7d02607daa
+STATEMENT: >>>
+PlantUML diagrams can be created inside of StrictDoc/RST markup as follows:
+
+.. code:: strictdoc
+
+    [TEXT]
+    STATEMENT: >>>
+    .. raw:: html
+
+        <pre class="plantuml">
+        @startuml
+        Alice -> Bob: Hello
+        @enduml
+        </pre>
+    <<<
+
+Markdown documents can use a fenced code block tagged ``plantuml`` instead:
+
+.. code:: markdown
+
+    ```plantuml
+    @startuml
+    Alice -> Bob: Hello
+    @enduml
+    ```
+
+Unlike Mermaid, PlantUML diagrams are not rendered in the browser: the
+diagram source is sent to a PlantUML server, which renders it to an SVG
+image. Because of this, PlantUML support is disabled by default and must be
+enabled explicitly by setting the ``plantuml_server_url`` option to a
+PlantUML server, for example the public server at
+``https://www.plantuml.com/plantuml``:
+
+.. code:: python
+
+    from strictdoc.core.project_config import ProjectConfig
+
+
+    def create_config() -> ProjectConfig:
+        config = ProjectConfig(
+            plantuml_server_url="https://www.plantuml.com/plantuml",
+        )
+        return config
+
+If ``plantuml_server_url`` is not set, ``pre.plantuml``/``` ```plantuml ```
+blocks are left as-is and no diagram is rendered.
+
+.. warning::
+
+    When using the public server, the text of every PlantUML diagram is sent
+    to a third party (plantuml.com) for rendering. Projects with
+    confidential diagram content should self-host a PlantUML server and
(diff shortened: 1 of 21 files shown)


stanislaw (approved):
Thanks a lot for the contribution, @bachp.

The only change I would request would be to enable a local instance of PlantUML server for the relevant e2e tests. The concern is that we could in theory run into flaky tests from time to time if for some reason the network access to the default PlantUML server would not work for some reason.

If you have a clear idea how to improve it, so that it works locally on GitHub CI Actions as well as locally for development, please go ahead and improve this right away. Otherwise, I need to think about the strategy, and we will implement this ourselves.

bachp:
@stanislaw I am not to familiar with github actions so I am not sure if this is possible. But idealy you could run the plantuml-server as a service container in CI, this is how I would do it in Gitlab.

Alternatively I just found out that plantuml can be compiled for the browser via WASM: https://github.com/plantuml/plantuml#-building-for-the-browser-teavm I need to check this out as this would simplify the rendering a lot and make it work the same way as mermaid.js does.

stanislaw:
> Alternatively I just found out that plantuml can be compiled for the browser via WASM: https://github.com/plantuml/plantuml#-building-for-the-browser-teavm I need to check this out as this would simplify the rendering a lot and make it work the same way as mermaid.js does.

This would be amazing 👍 Please let us know if it is doable.

stanislaw:
@bachp did you consider https://www.npmjs.com/package/@plantuml/core? It looks like this is what would work well for us.

bachp:
@stanislaw I didn't see the NPM package yet. I built it manually. What I don't know is how the vendoring policy is in this project. Would I just put a copy of the two .js files, taken from the npm package, into this project. Or is there a process for this?

stanislaw:
> *bachp* left a comment (strictdoc-project/strictdoc#3156)
> <https://github.com/strictdoc-project/strictdoc/pull/3156#issuecomment-5461924820>
>
> @stanislaw <https://github.com/stanislaw> I didn't see the NPM package
> yet. I built it manually. What I don't know is how the vendoring policy is
> in this project. Would I just put a copy of the two .js files, taken from
> the npm package, into this project. Or is there a process for this?

We need to check this more but I guess the npm package sort of assumes such
a use case. Let us know if you find out more. I should be able to get to
this tomorrow.

stanislaw:
@bachp It looks like the [@plantuml/core](https://www.npmjs.com/package/@plantuml/core) is MIT: https://github.com/plantuml/plantuml/blob/master/LICENSES.md. I believe we can inline it no problem. Or you would still see any issues?

bachp:
@stanislaw No I don't see an issue with the license or the vendoring per se. I was wondering more about how the vendored libraries (mermaid, mathjax, plantuml, ...) would be kept up to date. So it's more a general question and not plantuml specifc.

stanislaw:
> @stanislaw No I don't see an issue with the license or the vendoring per se. I was wondering more about how the vendored libraries (mermaid, mathjax, plantuml, ...) would be kept up to date. So it's more a general question and not plantuml specifc.

We don't have a special policy for this but we could set one up or even set up some automation to support this.

So far, it has been a couple of times that a user would discover a new feature release in one of the JS libraries and then they would contribute an upgrade.

stanislaw:
On this topic: StrictDoc development has several other periodic things that have to be done and we sometimes miss to perform the updates on time proactively:

- Updating specific docs
- Updating all sorts of Pip packages (for those that pinned)
- Synchronizing related repositories that are non linked as submodules but depend on StrictDoc's progress (Examples, Templates, strictdoc.tmLanguage, and some other).

Do you know of any best practices related to this?

PR #3162, opened 2026-08-29
feat: vendor PlantUML JS file, remove server-based code

stanislaw:
Follow-up to #3156


stanislaw:
cc @bachp I am investigating some issues with printing PlantUML to PDF. Otherwise, it seems to be working.

strictdoc/features/html2pdf/templates/features/html2pdf/index.jinja
```diff
@@ -18,29 +18,50 @@
   {%- endif -%}
```
stanislaw, line 18: @mettta this is the file with most drama in this PR.

PR #3173, opened 2026-09-05
Static check test cases

Robotics010:


Robotics010:
Have opened not on purpose, closing

PR #3212, opened 2026-09-15
Chrome Binary Pinning

gewlar:
Wires through the `--chrome-binary` argument added to html2pdf in https://github.com/strictdoc-project/html2pdf4doc_python/pull/105.
Adds two tests to check argument pick up.

Also adjusted the CI target from "master" to "main" to run them on our repository. If this is undesired I will revert those changes.


stanislaw:
> Also adjusted the CI target from "master" to "main" to run them on our repository. If this is undesired I will revert those changes.

These are very old artifacts that we no longer use and I missed to remove them long time ago. Instead of changing to main, I would remove them altogether as we don't want to double test the main branch after merges.

gewlar:
Do you mean removing those CI jobs as a whole or just the branch trigger?

stanislaw:
Just the branch triggers but I would even say just remove them from this changeset. I will review those and fix them separately.

stanislaw:
Please fix the linting issues and rebase/squash this branch to just one commit.

gewlar:
> Please fix the linting issues and rebase/squash this branch to just one commit.

Weird that conventional commits complains about reverted commits. Anyway rebased and squashed into a single commit.

stanislaw:
> Weird that conventional commits complains about reverted commits. Anyway rebased and squashed into a single commit.

This checker is project-specific and hand-written so it is expected behavior. We could extend it to support the reverts but there was simply no reason to do that just yet.

